PR #2, opened 2017-06-26
fixture middleware

niktrix:


fixture.js
```diff
@@ -0,0 +1,25 @@
+module.exports = fixture
+
+function fixture(staticResponses){
+  staticResponses = staticResponses || {}
+
+  return (req, res, next, end) => {
+    const self = this
+    staticResponses = staticResponses || {}
+    self.staticResponses = staticResponses
+    var staticResponse = self.staticResponses[req.method]
```
kumavis: prefer `const`

fixture.js
```diff
@@ -0,0 +1,25 @@
...
+      // async function
+    if ('function' === typeof staticResponse) {
+    staticResponse(req, next, end)
```
kumavis: missing `res` argument

fixture.js
```diff
@@ -0,0 +1,25 @@
+module.exports = fixture
+
+function fixture(staticResponses){
+  staticResponses = staticResponses || {}
+
+  return (req, res, next, end) => {
+    const self = this
```
kumavis: dont need `self` when using arrow fn's `=>`

test/fixture.js
```diff
@@ -0,0 +1,43 @@
...
+        t.ok(res, 'Has response')
+        t.equal(res.result, true, 'Response result is correct.')
+    }
+
+    function secondReqResponse(err, res) {
+        res2 = res
+        t.notOk(err, 'No error in response')
+        t.ok(res, 'Has response')
+        t.equal(res.result, '0x00', 'Response result is correct.')
+        t.end()
```
kumavis: should wrap both requests in an async/parallel or async/series so you can be sure both are done before closing the test

test/fixture.js
```diff
@@ -0,0 +1,43 @@
...
+
+test('fixture', (t) => {
+
+    const engine = new JsonRpcEngine()
+    let res1, res2
+
+    engine.push(fixture({
+        net_listening: true,
+        eth_hashrate: '0x00',
+        eth_mining: false,
```
kumavis: tests only cover the staticResponse usage

PR #5, opened 2018-09-13
Pass request object to getAccount calls

bitpshr:
This pull request passes the request object to `getAccount` calls, allowing for downstream origin checking before responding (re: EIP-1102.)


sema0703:
> This pull request passes the request object to `getAccount` calls, allowing for downstream origin checking before responding (re: EIP-1102.)

Ok

wallet.js
ALIR86, line 1: Solve bugs to uninstall extension browser in. Which default is best for I use of extension browser please provider  you concern me what errorr

PR #8, opened 2018-12-05
Support ERC191 v0 signatures

NoahZinsmeister:
Companion PR for [#5884 on `metamask-extension`](https://github.com/MetaMask/metamask-extension/pull/5884) adding support for [ERC191 v0](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-191.md) signatures via a new `eth_signTypedData_v0` RPC call.


kumavis:
@NoahZinsmeister thanks

PR #10, opened 2019-04-04
Remove babel

radotzki:
I suggest removing Babel because:
1. Babel is not needed in this package
2. It causing the build of [web3-provider-engine](https://github.com/MetaMask/web3-provider-engine) to fail


rekmarks (approved):
Our apologies for the delay.

PR #13, opened 2019-08-27
Fix interal error handling

Gudahtt:
`rpcErrors.internal` expects a string as the first argument, with arbitrary data being the optional second argument. Passing in a non-string first argument results in the error being shadowed by an error from eth-json-rpc-errors ('Error: "message" must be a nonempty string')


miranda76:
competed

Gudahtt:
@rekmarks This problem predates the changes you made, but thanks for the review!

PR #16, opened 2019-10-03
Update wallet.js

logvik:
We need it for https://github.com/MetaMask/metamask-extension/issues/1190

### DIFF
--- a/wallet.js
+++ b/wallet.js
@@ -206,34 +206,12 @@ function createWalletMiddleware(opts = {}) {
     // non-standard "extraParams" to be appended to our "msgParams" obj
     const extraParams = req.params[2] || {}
 
-    // We initially incorrectly ordered these parameters.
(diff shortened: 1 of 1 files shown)


wallet.js
```diff
@@ -196,6 +198,45 @@ function createWalletMiddleware(opts = {}) {
...
+      warning += `and has been corrected automatically. `
+      warning += `Please switch this param order for smooth behavior in the future.`
+      res.warning = warning
+
+      address = firstParam
+      message = secondParam
+    } else {
+      message = firstParam
+      address = secondParam
+    }
```
danfinlay: We shouldn't need this large block, since it's our first time defining this method. This was legacy logic for an old incorrect implementation of another signing method.

logvik:
That block is removed. Please review.

danfinlay:
Published as `eth-json-rpc-middleware@4.3.0`

PR #17, opened 2019-10-27
Update wallet.js

logvik:
Added new RCP method "encryption_public_key". It needs for getting the public key that will be used in the encrypt/decrypt feature. This key usually will be got via `nacl` library.
We need it for https://github.com/MetaMask/metamask-extension/issues/1190


logvik:
@danfinlay can we merge it?

PR #29, opened 2020-06-18
Normalize and require keyholder address

rekmarks:
### Terminology

What I call the "keyholder" address has previously been referred to as the "from" address and "sender" address. "from" and "sender" are misnomers, because the address simply needs to refer to an account controlled by MetaMask, whose private key is needed to perform some operation. Sometimes, this address is the recipient of a message. Other times, it is the sender/signer of a message.

Thus, "keyholder."

### Summary of Changes

- Require keyholder addresses wherever they are required in practice
- Validate _and_ normalize keyholder addresses whenever they are provided

### Changes in Detail

- Require keyholder addresses wherever they are required in practice
- Validate _and_ normalize keyholder addresses whenever they are provided
  - `validateSender` has been renamed to `validateAndNormalizeKeyholder`, and fulfills both purposes
  - Previously, the address would only be validated if provided. A valid address is now required.
  - Previously, the address would be normalized (i.e. lowercased) during validation without being overwritten. We now pass the normalized address to the extension if validation succeeds.
    - Note that we don't overwrite the parameters themselves; we only normalize the address value we use internally to identify the keyholder account.
- Update tests to reflect the above changes. All tests pass.

This PR will be followed by standardization PRs and a major version bump.

### DIFF
--- a/wallet.js
+++ b/wallet.js
@@ -2,9 +2,7 @@ const createAsyncMiddleware = require('json-rpc-engine/src/createAsyncMiddleware
 const createScaffoldMiddleware = require('json-rpc-engine/src/createScaffoldMiddleware')
 const sigUtil = require('eth-sig-util')
 
-module.exports = createWalletMiddleware
-
-function createWalletMiddleware(opts = {}) {
+module.exports = function createWalletMiddleware(opts = {}) {
   // parse + validate options
   const getAccounts = opts.getAccounts
   const processTypedMessage = opts.processTypedMessage
@@ -42,8 +40,7 @@ function createWalletMiddleware(opts = {}) {
   //
 
   async function lookupAccounts(req, res) {
-    const accounts = await getAccounts(req)
-    res.result = accounts
+    res.result = await getAccounts(req)
   }
 
   async function lookupDefaultAccount(req, res) {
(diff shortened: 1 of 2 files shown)


wallet.js
```diff
@@ -176,73 +191,89 @@ function createWalletMiddleware(opts = {}) {
...
-    await validateSender(secondParam, req)
     res.result = await processDecryptMessage(msgParams, req)
   }
 
   //
   // utility
   //
 
-  async function validateSender(address, req) {
-    // allow unspecified address (allow transaction signer to insert default)
```
Gudahtt, line 233: This seems to no longer be allowed. I'm not entirely sure yet which methods had allowed the "sender" address to be omitted, but requiring it all of a sudden seems like a breaking change.

rekmarks: I was trying to figure out whether we actually inserted a "default" as the comment suggests but couldn't find any examples. I'm going to run some experiments and see if you can request any of these RPC methods without specifying a keyholder address.

rekmarks: After testing `develop` and prod, I can confirm that we reject any request without a keyholder address for the wallet middleware RPC methods that require a keyholder address.

This includes `eth_sendTransaction`, where a default `from` address is injected by `web3` in our test dapp. We do not fallback to a default address in prod or `develop`.

Similarly, all signature methods fail in production if a keyholder address is not specified. Our UI handles this very poorly, so rejecting them in this middleware layer will circumvent that problem entirely.

PR #34, opened 2020-06-23
Update wallet middleware errors

rekmarks:
- Add `eth-rpc-errors`
- Replace wallet middleware errors with `eth-rpc-errors` where appropriate
  - Update tests
- Update fetch middleware to use `eth-rpc-errors` instead of `eth-json-rpc-errors`, which is deprecated
- Remove `eth-json-rpc-errors`

### DIFF
--- a/fetch.js
+++ b/fetch.js
@@ -1,6 +1,6 @@
 const fetch = global.fetch || require('fetch-ponyfill')().fetch
 const url = require('url')
-const { errors: rpcErrors } = require('eth-json-rpc-errors')
+const { ethErrors } = require('eth-rpc-errors')
 const btoa = require('btoa')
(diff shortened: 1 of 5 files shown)


fetch.js
```diff
@@ -73,10 +73,16 @@ function checkForHttpErrors (fetchRes) {
...
-    throw rpcErrors.internal(`Non-200 status code: '${fetchRes.status}'`, body)
+    throw ethErrors.rpc.internal({
+      message: `Non-200 status code: '${fetchRes.status}'`,
+      data: body,
+    })
   }
   // check for rpc error
-  if (body.error) throw rpcErrors.internal(body.error.toString(), body.error)
+  if (body.error) throw ethErrors.rpc.internal({
+    message: body.error.toString(),
```
Gudahtt, line 83: This parameter should be dropped (see #20), but this can be done in a separate PR.

PR #52, opened 2020-09-14
v5.0.3

whymarrh:
This PR represents v5.0.3, the next patch release.

This release mainly includes #51.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "eth-json-rpc-middleware",
-  "version": "5.0.2",
+  "version": "5.0.3",
   "main": "block-ref.js",
   "directories": {
     "test": "test"


Gudahtt (approved):
LGTM!

In the future I'd suggest using a different branch name for release branches. Using a branch with a matching name to the tag makes it difficult to reference them using the git CLI, as the reference becomes ambiguous. I use `release-[version]` but it doesn't really matter what you use as long as it's different from the tag.

whymarrh:
The tag doesn't exist until the branch is merged though

whymarrh:
> A new version of the package `eth-json-rpc-middleware` (5.0.3) was published at 2020-09-14T20:38:24.692Z from
[redacted]. The shasum of this package was `85ba10f0a1efd2523123823a934a7f3a9e41268f`.

PR #56, opened 2020-09-22
Remove VM middleware, dependencies

rekmarks:
We don't use this middleware anywhere, and its dependencies are outdated by years, which means it's been pretty useless for years.


kumavis:
publish as a breaking change +1

kumavis (approved):

PR #61, opened 2020-12-09
Standardize linting

### LINKED ISSUE #46: Use shared ESLint config
This project should used the shared ESLint config

NiranjanaBinoy:
The purpose of this PR is to introduce eslint to eth-json-rpc-middleware.

Closes #46


.eslintrc.js
```diff
@@ -0,0 +1,29 @@
...
+  ignorePatterns: [
+    '!.eslintrc.js',
+    'node_modules/',
+  ],
+  overrides: [
+    {
+      files: ['*js'],
+      rules: {
+        'consistent-return': 'off',
+        'node/callback-return': 'off',
```
NiranjanaBinoy, line 25: Is adding these rules the correct approach?

src/block-ref-rewrite.js
```diff
@@ -9,17 +9,24 @@ function createBlockRefRewriteMiddleware (opts = {}) {
...
+    if (blockRef === undefined) {
+      blockRef = 'latest'
+    }
+    if (blockRef !== 'latest') {
+      return next()
+    }
     // rewrite blockRef to block-tracker's block number
     const latestBlockNumber = await blockTracker.getLatestBlock()
+    // eslint-disable-next-line require-atomic-updates
     req.params[blockRefIndex] = latestBlockNumber
```
NiranjanaBinoy, line 30: The error claimed this will cause a race around the condition, but on studying the code it does not seem like a likely condition. Hence I disabled `require-atomic-updates`. Is there any alternate solution?

src/fetch.js
```diff
@@ -57,6 +59,7 @@ function createFetchMiddleware ({ rpcUrl, originHttpHeaderKey }) {
 
 function checkForHttpErrors (fetchRes) {
   // check for errors
+  // eslint-disable-next-line default-case
   switch (fetchRes.status) {
```
NiranjanaBinoy, line 63: I could not figure out a default case to define for this case, so I disabled the error for this lint. Is there a default case that can be definedhere?

src/cache-utils.js
```diff
@@ -147,6 +147,7 @@ function cacheTypeForPayload (payload) {
     case 'shh_getFilterChanges':
     case 'shh_getMessages':
     case 'test_neverCache':
+    default:
       return 'never'
   }
```
NiranjanaBinoy, line 152: This switch case for cache method had `default-case` error. Since there are multiple cases are using `never` cache, I made an assumption that the `default` condition will be `never`. Is there any other value that can be defined?

src/wallet-parity.js
```diff
@@ -52,7 +52,8 @@ function createParityWalletMiddleware({ walletMiddleware }){
     })
   }
 
-  async function checkRequest (req, res) {
+  async function checkRequest (_req, res) {
+    // eslint-disable-next-line no-undef
     const result = parityRequests[reqId]
```
NiranjanaBinoy, line 57: Even though I disabled this `no-undef` error, the `reqId` variable in this line does not have a definition or value, and also I was unable to track down any usage of `createParityWalletMiddleware`.

.eslintrc.js
rekmarks: I think we should keep these rules enabled, and fix all of the no doubt many violations 😅 

If it's too painful to fix in tests, feel free to disable the rules for tests for now. Ideally we should just fix them everywhere.

src/block-ref-rewrite.js
rekmarks: You made the right call here. If we use `json-rpc-engine` correctly, which AFAIK we do, the only code that will modify `req` while this function is executing, is this function.

src/wallet-parity.js
rekmarks: On review, we should delete this file and all references to it. It is not referenced anywhere in the extension codebase, and clearly it's in a bad state anyway.

rekmarks (changes requested):
I responded to all comments, and made some requests for changes.

src/wallet-parity.js
NiranjanaBinoy: I don't think this file is referenced anywhere in this repo either. I did check the extension and could not find any reference there either. Should I remove this file as a part of this PR or a separate PR??

rekmarks: I wanted to say just do it here, but we should be good and do it in a separate PR, as I've done here: https://github.com/MetaMask/eth-json-rpc-middleware/pull/63

src/cache-utils.js
```diff
@@ -69,7 +69,9 @@ function blockTagParamIndex (payload) {
...
 }
 
+// eslint-disable-next-line consistent-return
 function cacheTypeForPayload (payload) {
+  // eslint-disable-next-line default-case
```
rekmarks, line 74: We should avoid disabling rules when we can, and I believe we resolve both rule violations without changing the function's behavior by adding the following default case:
```javascript
default:
  return undefined
```

rekmarks: We can then default to `'never'` in a follow-up, as discussed.

src/fetch.js
```diff
@@ -79,14 +83,17 @@ function parseResponse (fetchRes, body) {
...
+    })
+  }
   // return successful result
   return body.result
 }
 
-function createFetchConfigFromReq({ req, rpcUrl, originHttpHeaderKey }) {
+function createFetchConfigFromReq ({ req, rpcUrl, originHttpHeaderKey }) {
+  // eslint-disable-next-line node/no-deprecated-api
   const parsedUrl = url.parse(rpcUrl)
```
rekmarks, line 97: We should fix this in a follow-up: https://github.com/MetaMask/eth-json-rpc-middleware/issues/64

src/cache-utils.js
rekmarks: Issue for the follow-up: https://github.com/MetaMask/eth-json-rpc-middleware/issues/65

rekmarks (changes requested):
Some comments, some requested changes.

rekmarks (approved):
LGTM!

PR #66, opened 2020-12-15
Default unknown methods to 'never' in block cache middleware 

### LINKED ISSUE #65: Default unknown methods to 'never' in block cache middleware
Unknown methods should never be cached, and classified as such.

NiranjanaBinoy:
Unknown methods should never be cached and classified as such.

closes #65


rekmarks (approved):
LGTM!

PR #67, opened 2020-12-15
Replace 'url' dependency with native URL global

### LINKED ISSUE #64: Replace 'url' dependency with native URL global
We should just use the native `URL` global whenever we deal with URLs. We should also avoid any deprecated methods of that API.

NiranjanaBinoy:
We should just use the native URL global whenever we deal with URLs.
closes #64


rekmarks (approved):
Nice!

PR #68, opened 2021-01-12
Typescript migration

### LINKED ISSUE #59: Migrating eth-json-rpc-middleware to Typescript

NiranjanaBinoy:
Migrating eth-json-rpc-middleware to typescript.

closes #59


src/block-cache.ts
```diff
@@ -0,0 +1,183 @@
...
+  async set(payload: JsonRpcRequest<string[]>, requestedBlockNumber: string, result: Record<string, unknown>): Promise<void> {
+    // check if we can cached this result
+    const canCache: boolean = this.canCacheResult(payload, result);
+    if (!canCache) {
+      return;
+    }
+    // set the value in the cache
+    const blockCache: Record<string, unknown> = this.getBlockCacheForPayload(payload, requestedBlockNumber);
+    const identifier: string|null = cacheUtils.cacheIdentifierForPayload(payload, true);
+    blockCache[(identifier as string)] = result;
```
rekmarks, line 66: I know that these lines preserve existing functionality, but this looks like a bug to me. If `identifier` is `null`, we should return early, just like we do in the `canCache` check above.

src/block-cache.ts
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  canCacheRequest(payload: JsonRpcRequest<string[]>,): boolean {
```
rekmarks, line 69: There's a stray comma after the `payload` param.

src/block-cache.ts
```diff
@@ -0,0 +1,183 @@
...
+    }
+    // can be cached
+    return true;
+  }
+
+  canCacheResult(payload: JsonRpcRequest<string[]>, result: Record<string, unknown>): boolean {
+    // never cache empty values (e.g. undefined)
+    if (emptyValues.includes((result as unknown) as string)) {
+      return false;
+    }
```
rekmarks, line 87: If the `result` can be something other than `Record<string, unknown>`, we should type it with an appropriate union type. After the `emptyValues.includes` type guard, TypeScript should be able to correctly infer that `result` is a `Record<string, unknown>`.

src/block-cache.ts
```diff
@@ -0,0 +1,183 @@
...
+      }
+    }
+    // otherwise true
+    return true;
+  }
+
+  // removes all block caches with block number lower than `oldBlockHex`
+  clearBefore(oldBlockHex: string): void {
+    // eslint-disable-next-line @typescript-eslint/no-this-alias
+    const self = this;
```
rekmarks, line 101: I think we can delete `self` variable. I can't see anywhere where `this` wouldn't have the correct reference. All the arrow functions are lexically bound to the parent scope, i.e. the `clearBefore` function.

src/block-cache.ts
```diff
@@ -0,0 +1,183 @@
...
+    perma: blockCache,
+    block: blockCache,
+    fork: blockCache,
+  };
+
+  return createAsyncMiddleware(async (req, res, next) => {
+    // allow cach to be skipped if so specified
+    if (((req as unknown) as Record<string, unknown>).skipCache) {
+      return next();
+    }
```
rekmarks, line 133: I think we can define an interface that supersets the interface expected by `createAsyncMiddleware` for the `req` param that we can cast to on L131, instead of doing current wordy and less specific cast.

src/block-cache.ts
```diff
@@ -0,0 +1,183 @@
...
+    // create new cache if necesary
+    if (!blockCache) {
+      const newCache: Record<string, unknown> = {};
+      this.cache[blockNumber] = newCache;
+      blockCache = newCache;
+    }
+    return blockCache;
+  }
+
+  async get(payload: JsonRpcRequest<string[]>, requestedBlockNumber: string): Promise<unknown> {
```
rekmarks, line 44: I think we can do better here than `Promise<unknown>`. We shouldn't have to cast the return value of this function where it's used.

src/fetch.ts
```diff
@@ -0,0 +1,183 @@
...
+
+function createTimeoutError(): EthereumRpcError<unknown> {
+  let msg = `Gateway timeout. The request took too long to process. `;
+  msg += `This can happen when querying logs over too wide a block range.`;
+  return ethErrors.rpc.internal({ message: msg });
+}
+
+function timeout(duration: number): Promise<unknown> {
+  return new Promise((resolve) => setTimeout(resolve, duration));
+}
```
rekmarks, line 183: This should have a return type of `Promise<ReturnType<setTimeout>>`.

src/block-cache.ts
rekmarks: That could then also be used as the return type for the `JsonRpcMiddleware` return value of `createBlockCacheMiddleware` (i.e. the function we're in).

src/block-cache.ts
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    // get exact block number
+    let requestedBlockNumber: string;
+    if (blockTag === 'earliest') {
+      // this just exists for symmetry with "latest"
+      requestedBlockNumber = '0x00';
+    } else if (blockTag === 'latest') {
+      // fetch latest block number
+      const latestBlockNumber: string = await blockTracker.getLatestBlock();
```
rekmarks, line 159: nit: We don't need to explicitly specify here; it will be inferred. There are sometimes DevX and potentially safety benefits to "unnecessarily" specifying types (especially with function/method return types), but I don't think this is such an example.

src/block-cache.ts
```diff
@@ -0,0 +1,183 @@
...
+    if (cacheResult === undefined) {
+      // cache miss
+      // wait for other middleware to handle request
+      // eslint-disable-next-line node/callback-return
+      await next();
+      // add result to cache
+      await strategy.set(req, requestedBlockNumber, (res.result as Record<string, unknown>));
+    } else {
+      // fill in result from cache
+      res.result = (cacheResult as Record<string, unknown>);
```
rekmarks, line 179: Per feedback elsewhere in this function, it would be great if we could do less `Record<string, unknown>` and casting here. Probably (also as suggested elsewhere), by manually defining some interfaces.

src/block-ref-rewrite.ts
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  return createAsyncMiddleware(async (req, _res, next) => {
+    const blockRefIndex: number|undefined = blockTagParamIndex(req);
+    // skip if method does not include blockRef
+    if (blockRefIndex === undefined) {
+      return next();
+    }
+    // skip if not "latest"
+    let blockRef: string = (req.params as string[])[blockRefIndex];
```
rekmarks, line 29: We can define an interface as follows to avoid this cast:
```typescript
interface BlockRefJsonRpcRequest extends JsonRpcRequest<string[]> {
  params: string[];
}
```

src/block-ref.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  return createAsyncMiddleware(async (req, res, next) => {
+    const blockRefIndex: number|undefined = blockTagParamIndex(req);
+    // skip if method does not include blockRef
+    if (blockRefIndex === undefined) {
+      return next();
+    }
+    // skip if not "latest"
+    let blockRef: string = (req.params as string[])[blockRefIndex];
```
rekmarks, line 39: See comment in `block-ref-rewrite` for interface to avoid this cast.

src/block-ref.ts
```diff
@@ -0,0 +1,60 @@
...
+import clone from 'clone';
+import pify from 'pify';
+import { blockTagParamIndex } from './cache-utils';
+
+// eslint-disable-next-line @typescript-eslint/no-var-requires,@typescript-eslint/no-require-imports
+const BlockTracker = require('eth-block-tracker');
+
+interface BlockRefMiddlewareOptions{
+  blockTracker?: typeof BlockTracker;
+  provider?: Record<string, unknown>;
```
rekmarks, line 15: We should create an interface for `provider` with the properties we expect. I think it's just a single function, `sendAsync`, with a `JsonRpcRequest` as its first parameter and an error-first callback as the second. It could be genericized, e.g.:
```typescript
interface Provider {
  sendAsync<T, U>: (JsonRpcRequest<T>, (err: Error, result: JsonRpcResponse<U>) => void) => void;
}
```

src/block-tracker-inspector.ts
```diff
@@ -0,0 +1,38 @@
...
+      return next();
+    }
+    // eslint-disable-next-line node/callback-return
+    await next();
+    // abort if no result or no block number
+    if (!res.result || !res.result.blockNumber) {
+      return undefined;
+    }
+    // if number is higher, suggest block-tracker check for a new block
+    const blockNumber: number = Number.parseInt((res.result as Record<string, string>).blockNumber, 16);
```
rekmarks, line 31: We should ensure that `res.result` is appropriately typed earlier in the function, if possible.

rekmarks (changes requested):
Wonderful work converting this massive package. This is a huge step forward to converting our internal dependency graph to TypeScript.

That said, there's still a fair amount of work to be done here, I think!

Going alphabetically through `src`, I got through `block-tracker-inspector.ts`. In general, I think we should use define more interfaces and use less `Record<string, unknown>` types and type casting generally (and especially `as unknown as SomethingElse`) throughout .

Gudahtt (commented):
The `main` entry in `package.json` should be updated to point at `./dist/block-ref.js`, so that the main package export continues to work.

src/cache-utils.ts
```diff
@@ -0,0 +1,137 @@
+import {
+  JsonRpcRequest,
+  PendingJsonRpcResponse,
+} from 'json-rpc-engine';
+import stringify from 'json-stable-stringify';
+import SafeEventEmitter from '@metamask/safe-event-emitter';
+
+export type Payload = Partial<JsonRpcRequest<string[]>>;
+export interface JsonRPCRequestToCache extends JsonRpcRequest<string[]>{
```
rekmarks, line 9: nit: casing:
```suggestion
export interface JsonRpcRequestToCache extends JsonRpcRequest<string[]>{
```

src/retryOnEmpty.ts
```diff
@@ -0,0 +1,109 @@
...
+// `<nil>` comes from https://github.com/ethereum/go-ethereum/issues/16925
+const emptyValues: (string|null|undefined)[] = [undefined, null, '\u003cnil\u003e'];
+
+interface RetryOnEmptyMiddlewareOptions{
+  provider?: SafeEventEmitterProvider;
+  blockTracker?: typeof BlockTracker;
+}
+
+function createRetryOnEmptyMiddleware(
+  opts: RetryOnEmptyMiddlewareOptions = {}
```
rekmarks, line 36: As noted elsewhere, `opts` should be destructured here, if possible. We can keep the default for now.

src/wallet.ts
```diff
@@ -0,0 +1,287 @@
...
+  processTransaction?: (txParams: ParamsObject, req: JsonRpcRequest<unknown>) => Promise<ParamsObject>;
+  processTypedMessage?: (msgParams: ParamsObject, req: JsonRpcRequest<unknown>, version: string) => Promise<ParamsObject>;
+  processTypedMessageV3?: (msgParams: ParamsObject, req: JsonRpcRequest<unknown>, version: string) => Promise<ParamsObject>;
+  processTypedMessageV4?: (msgParams: ParamsObject, req: JsonRpcRequest<unknown>, version: string) => Promise<ParamsObject>;
+}
+
+export = createWalletMiddleware;
+
+function createWalletMiddleware(
+  opts: WalletMiddlewareOptions
```
rekmarks, line 30: As noted elsewhere, `opts` should be destructured here.

src/wallet.ts
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    const message: string = (req.params as string[])[0];
+    const address: string = await validateAndNormalizeKeyholder((req.params as string[])[1], req);
+    const version = 'V1';
+    const extraParams: ParamsObject = (req.params as ParamsObjectList)[2] || {};
+    const msgParams: ParamsObject = Object.assign({}, extraParams, {
+      from: address,
+      data: message,
+    });
```
rekmarks, line 127: For stylistic purposes (and arguably readability), it would be nice to convert the `Object.assign` calls in this file to object spreads, like so:
```suggestion
    const msgParams: ParamsObject = {
      ...extraParams,
      from: address,
      data: message,
    });
```

rekmarks (changes requested):
Excellent work! It'd be nice to improve the typing of `ParamsObject` in `wallet.ts`, however we don't see to do any/sufficient runtime type checking anyway, so those types would be useless or even dangerous in practice.

I left a lot of inline comments, but this is shaping up nicely!

src/block-cache.ts
```diff
@@ -0,0 +1,193 @@
...
+      const newCache: BlockCache = {};
+      this.cache[blockNumber] = newCache;
+      blockCache = newCache;
+    }
+    return blockCache;
+  }
+
+  async get(payload: Payload, requestedBlockNumber: string): Promise<Block|undefined> {
+    // lookup block cache
+    const blockCache: BlockCache = this.getBlockCacheForPayload(payload, requestedBlockNumber);
```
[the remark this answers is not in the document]
NiranjanaBinoy: In `getBlockCacheForPayload`, if a block is not found it is adding a new empty block. I thought that might make the `blockCache === undefined` redundant!!

src/block-cache.ts
```diff
@@ -0,0 +1,193 @@
...
+    if (!canCacheResult) {
+      return;
+    }
+    const identifier: string|null = cacheIdentifierForPayload(payload, true);
+    if (!identifier) {
+      return;
+    }
+    // set the value in the cache
+    const blockCache: BlockCache = this.getBlockCacheForPayload(payload, requestedBlockNumber);
+    blockCache[identifier] = result;
```
rekmarks, line 76: I originally asked that we add an early return here, which we should, but to be good, we should do that in a follow-up. We should do this for now, and I'll create an issue:
```suggestion
    // set the value in the cache
    const blockCache: BlockCache = this.getBlockCacheForPayload(payload, requestedBlockNumber);
    // TODO: Add early return if the identifier is falsy to avoid this cast
    blockCache[identifier as any] = result;
```

rekmarks: Issue: #73

src/block-cache.ts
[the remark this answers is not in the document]
rekmarks: Ah, you're right. Never mind!

src/block-cache.ts
```diff
@@ -0,0 +1,193 @@
...
+    const cacheResult: Block|undefined = await strategy.get(req, requestedBlockNumber);
+    if (cacheResult === undefined) {
+      // cache miss
+      // wait for other middleware to handle request
+      // eslint-disable-next-line node/callback-return
+      await next();
+
+      if (res.result === undefined) {
+        return undefined;
+      }
```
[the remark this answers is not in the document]
NiranjanaBinoy: To avoid typecasting of `res.result` on `await strategy.set(req, requestedBlockNumber, res.result);`
I couldn't decide which is the better approach!!

src/block-cache.ts
```diff
@@ -0,0 +1,193 @@
...
+    if (blockTag === 'pending') {
+      return false;
+    }
+    // can be cached
+    return true;
+  }
+
+  canCacheResult(payload: Payload, result: Block): boolean {
+    // never cache empty values (e.g. undefined)
+    if (emptyValues.includes((result as unknown) as string)) {
```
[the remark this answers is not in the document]
NiranjanaBinoy: `unknown` does not work, so I am going with `any`

src/block-cache.ts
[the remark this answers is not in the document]
rekmarks: In this case we should cast, because `BlockCacheStrategy.set` calls `canCacheResult`, which checks if `res.result` is one of the "empty" values. This package otherwise assumes that the RPC endpoints return correct data, so there are no more runtime type checks to perform.

I made the change in 3a4c94f, to decrease turnaround.

rekmarks (approved):
Let's gooo! 😄

PR #69, opened 2021-02-02
Update `ganache-core`

Gudahtt:
The latest version of `ganache-core` is now used. The resolutions used previously to address security advisories are no longer necessary.


rekmarks (approved):
Very nice.

Issue #73, opened 2021-03-11
Block cache "set" method is missing early return for missing identifier


rekmarks:
In the `block-cache` middleware, the `set` function should return early if no identifier is found. The current behavior may attempt to set values for keys of `null` or `undefined`.

PR #75, opened 2021-03-12
Block cache "set" method returns early for a null/ undefined identifier.

NiranjanaBinoy:
fixes : #73 

In the block-cache middleware, the set function is updated to return early if the identifier is not found (null/undefined)


rekmarks (changes requested):
We should move the call to `getBlockCacheForPayload` beneath the early return, like you did in your original implementation!

rekmarks (approved):

PR #76, opened 2021-03-12
Robustify providerFromEngine callback type check

### LINKED ISSUE #72: Improve runtime type check in providerFromEngine send function
Currently, it checks that a truthy `callback` param is provided, but really it should check that that it's `typeof 'function'`.

rekmarks:
Fixes #72 

Checks that the callback provided to the `send` method of the provider object returned by `providerFromEngine` checks that the callback is a function, rather than just truthy.


NiranjanaBinoy (approved):
LGTM

PR #77, opened 2021-03-12
Adding interfaces msgParams in wallet.ts

### LINKED ISSUE #74: The msgParams parameters in wallet.ts should be typed better
Pending #68 

The `msgParams` parameters in `wallet.ts` should be manually typed with interfaces as opposed to `ParamsObject`, which is just an alias for `Record<string, unknown>`.

NiranjanaBinoy:
- Added new interfaces and updated the corresponding usage
- Removed the aliases ParamsObject and ParamsObjectList with `Record<string,  unknown>`, and `Record>string, unknown>[]`

Closes #74

### DIFF
--- a/src/wallet.ts
+++ b/src/wallet.ts
@@ -21,7 +21,7 @@ interface TypedMessageParams extends MessageParams {
   version: string;
 }
 
-interface WalletMiddlewareOptions{
+interface WalletMiddlewareOptions {
   getAccounts: (req: JsonRpcRequest<unknown>) => Promise<string[]>;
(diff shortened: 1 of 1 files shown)


src/wallet.ts
```diff
@@ -9,19 +9,27 @@ import * as sigUtil from 'eth-sig-util';
...
+interface TransactionParams {
+  from: string;
+}
+
+interface MessageParams extends TransactionParams{
+  data: string;
+}
+interface TypedMessageParams extends MessageParams{
+  version: string;
+}
```
rekmarks, line 21: interface MessageParams extends TransactionParams {
  data: string;
}

interface TypedMessageParams extends MessageParams {
  version: string;
}

rekmarks (approved):
LGTM!

PR #78, opened 2021-03-23
Add config to exclude node-fetch in browser builds

### LINKED ISSUE #71: Remove node-fetch from production dependencies
The fetch middleware currently falls back to `node-fetch` if `global.fetch` is `undefined`. I want to say that dealing with that is not our problem, and we should get rid of it. I'm planning a breaking change for this package anyway.

rekmarks:
Fixes #71


NiranjanaBinoy (approved):
LGTM

PR #80, opened 2021-03-25
remove unused dependencies

brad-decker:
went to upgrade ethereumjs-util and then realized its not being used. Also eth-query was a similar way so i removed both.


ryanml (commented):
looks good as i think these are still both sub-deps of other dependencies, but just wanted to double check to make sure there were no needed modifications to `yarn.lock`

brad-decker:
i swore i ran yarn remove. Thanks for pointing this out @ryanml !

brad-decker:
weird, well did the process again and it still doesn't result in a lockfile change

ryanml (approved):
Hm, the change could very well just not be needed. Thanks for double checking on it!

Gudahtt:
If another dependency asks for precisely the same range of each of these packages, I'd expect there to be no changes. Not especially common, but I guess that's what happened here!

PR #81, opened 2021-03-25
Refactor exports

rekmarks:
- Use named exports only
- Add `index.ts` that re-exports everything except `cache-utils`, which is only used internally
- Move `cache-utils` to separate `utils` folder to emphasize that it's different from the rest of the source files


NiranjanaBinoy (approved):
LGTM

PR #82, opened 2021-03-25
Update eslint, CI, TypeScript configs

rekmarks:
- Update `.eslintrc.js` per `metamask-module-template`
  - Remove assumption that we're in the DOM
- Update `tsconfig.json`
  - Remove assumption that we're in the DOM
  - Add inline reference to DOM lib `src/fetch.ts` because it needs it
- Update CircleCI config
  - This was necessary because with the updated config, `eslint` doesn't accept unresolved imports in the test directory
- Add `@metamask/eslint-config@latest`, `@types/node`
- Run `prettier` on all `src` and `test` files, then `yarn lint:fix`
  - All changes to `test/` and `src/` files are programmatic _except_ the aforementioned DOM lib reference, which was added manually

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,20 +3,20 @@ version: 2.1
 workflows:
   build-test:
     jobs:
-      - prep-deps
+      - prep-deps-build
       - test-lint:
           requires:
-            - prep-deps
+            - prep-deps-build
       - test-unit:
           requires:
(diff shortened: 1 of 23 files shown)


NiranjanaBinoy (approved):
lgtm

brad-decker (approved):
LGTM!

PR #83, opened 2021-03-25
Fix DOM TypeScript library usage

rekmarks:
This removes the triple-slash directive to use the TypeScript DOM lib and puts it in `tsconfig.json` where it belongs. I thought the triple-slash directive add the given lib on a per-file basis, but it adds it for the entire project. Lame!

This also allowed us to remove an override from the eslint config.


NiranjanaBinoy (approved):
LGTM

PR #84, opened 2021-03-25
Ignore btoa dependency in browser environments

rekmarks:
Ensures that appropriately configured bundlers ignores the `btoa` dependency in favor of the browser implementation in browser environments.


NiranjanaBinoy (approved):
LGTM

PR #91, opened 2021-04-06
Add dependabot to keep modern dependencies

darkwing:


Gudahtt (changes requested):
This needs to be updated to match our newer Dependabot config, which only updates `@metamask` dependencies.

rekmarks:
Superseded by #105.

PR #96, opened 2021-07-26
Add support for eth_signTransaction so that external key stores can be used for signing

prem-baranwal:
This pull request is to support two minor features in the eth-json-rpc-middleware library-

1. Add support for _SafeEventEmitterProvider::send_ interface to be able to use **Web3.Providers.HttpProvider or Web3.Providers.WebsocketProvider**. To be able to use code like `this.engine.push(providerAsMiddleware(new Web3.Providers.WebsocketProvider));`, providerFromEngine need to expose send method with callback. Currently its not supported. (Files changes : src/providerAsMiddleware.ts, src/providerFromEngine.ts and src/utils/cache.ts). Since SafeEventEmitterProvider::send interface was not being used anywhere in the repo, the code changes are minimal.
2. Add support for **eth_sendRawTransaction** so that a transaction signed externally (using online key stores like Azure key vault) can be sent to the geth node.  (Files changes : src/wallet.ts)

### DIFF
--- a/src/wallet.ts
+++ b/src/wallet.ts
@@ -35,15 +35,15 @@ interface WalletMiddlewareOptions {
     msgParams: MessageParams,
     req: JsonRpcRequest<unknown>
   ) => Promise<Record<string, unknown>>;
-  processEthSignTransactionMessage?: (
-    txParams: TransactionParams,
-    req: JsonRpcRequest<unknown>
-  ) => Promise<Record<string, unknown>>;
   processPersonalMessage?: (
     msgParams: MessageParams,
     req: JsonRpcRequest<unknown>
   ) => Promise<Record<string, unknown>>;
   processTransaction?: (
+    txParams: TransactionParams,
+    req: JsonRpcRequest<unknown>
+  ) => Promise<Record<string, unknown>>;
+  processSignTransaction?: (
     txParams: TransactionParams,
     req: JsonRpcRequest<unknown>
   ) => Promise<Record<string, unknown>>;
@@ -69,9 +69,9 @@ export function createWalletMiddleware({
   processDecryptMessage,
   processEncryptionPublicKey,
   processEthSignMessage,
-  processEthSignTransactionMessage,
   processPersonalMessage,
   processTransaction,
+  processSignTransaction,
   processTypedMessage,
   processTypedMessageV3,
   processTypedMessageV4,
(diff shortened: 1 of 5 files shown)


Gudahtt:
Thanks for the contribution! We are definitely interested in updating `providerAsMiddleware` to support [EIP-1193](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1193.md)-compliant providers, which would include the `HttpProvider` offered by `ethers` (I'm guessing that's what you're referring to here?). But both `send` and `sendAsync` are deprecated in EIP-1193. This should be using the newer `request` method.

For the second point, I'm guessing that you meant to say `eth_signTransaction`? We don't support signing a transaction without broadcasting it yet on the extension, but we have considered it. I'll check with the team first, but I think this is something we could consider accepting in this library regardless. I would ask that you create a separate PR for each of these changes though, as they are quite distinct from each other.

prem-baranwal:
I have reverted the provider related changes. Current pull request is now only for eth_signTransaction addition. 

Also, if you can point me to some EIP-1193 provider implementation, i can check and tweak my other changes. I looked into [web3js](https://github.com/ChainSafe/web3.js) as well as [ethersjs](https://github.com/ethers-io/ethers.js) but they still seem to be using the older methods. If there are no `request` implementations, it might be a good idea to include `send` for the time being.

src/wallet.ts
```diff
@@ -160,6 +166,25 @@ export function createWalletMiddleware({
     res.result = await processEthSignMessage(msgParams, req);
   }
 
+  async function ethSignTransaction(
```
Gudahtt, line 169: Could you move this function back up to the "transaction signatures" section?

Gudahtt (changes requested):
Thanks! This code all looks good to me. I left one comment about moving it into the right section, but also could you add unit tests for this new method?

Since it's functional equivalent to `sendTransaction` apart from the final function it calls, you should be able to copy the `transactionTest` helper and the `sendTransaction` tests when writing them.

Gudahtt:
>Also, if you can point me to some EIP-1193 provider implementation, i can check and tweak my other changes. I looked into web3js as well as ethersjs but they still seem to be using the older methods. If there are no request implementations, it might be a good idea to include send for the time being.

My apologies, I thought the Ethers providers had an EIP-1193-like interface. What you added originally makes sense to me, except that it might be better to call it something like `ethersProviderAsMiddleware`, as the expectations are specific to ethers.

src/wallet.ts
prem-baranwal: I have moved this function next to sendTransaction/processTransaction and added the testcases.

prem-baranwal:
Updated as per the comment. Also will create a separate pull request for provider `send` method support.

Gudahtt (approved):
LGTM!

PR #97, opened 2021-07-29
add support for send call with callback so that Web3 Providers can be used

prem-baranwal:
Creating a pull request off from pull request #96 as both the requests are independent changes. This request is to add support for adding Web3 providers as middleware which requires calling `send` instead of `sendAsync`. The existing `send` interface is missing the callback argument.
Please refer to pull request #96 for original discussion.


Gudahtt (approved):
LGTM!

PR #104, opened 2021-10-31
chore: replace json-stable-stringify with safe-stable-stringify

BridgeAR:
This is a performance improvement as the latter is actually faster.

json-stable-stringify x 13,870 ops/sec ±0.72% (94 runs sampled)
safe-stable-stringify x 30,367 ops/sec ±0.39% (96 runs sampled)

The only difference is that objects with circular reference are from
now on also accepted instead of throwing an error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+- Replace `fast-stable-stringify` dependency with `safe-stable-stringify`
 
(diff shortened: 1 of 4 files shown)


rekmarks:
Thank you for your contribution! This needs to be rebased, but in any event, I don't believe that we want to accept circular references in any way. Would you be willing to add that as a feature to `safe-stable-stringify`?

BridgeAR:
@rekmarks thanks for the feedback. I went ahead and implemented such option and updated the PR accordingly.

Gudahtt:
Hey @BridgeAR , sorry for the delay! I can review this again once the conflicts are resolved

BridgeAR:
@Gudahtt thanks, I just rebased the PR.

Gudahtt (approved):
LGTM!

PR #108, opened 2021-11-23
fetch - dont override server returned error code and message

kumavis:


Elyx0:
Do you have a screenshot or logs of the output of the same error before / after this?

sambacha:
still causing issues ser

PR #109, opened 2021-12-07
fix: add query parameters to normalized url

islishude:
https://github.com/MetaMask/metamask-extension/issues/12995


Gudahtt (approved):
Great catch! I see that in v6 we used `path` but then we switched to `pathname` in #67 which excludes the query string.

Gudahtt:
The "Require Additional Reviewer" failures are caused by a bug in that action. I will update it in a separate PR.

georgewrmarshall (commented):
lgtm

PR #116, opened 2022-03-22
Added getAccounts suppressUnauthorized param

shanejonas:
rel https://github.com/MetaMask/metamask-extension/pull/14126


src/wallet.ts
```diff
@@ -379,7 +382,7 @@ export function createWalletMiddleware({
   ): Promise<string> {
     if (typeof address === 'string' && address.length > 0) {
       // ensure address is included in provided accounts
-      const accounts: string[] = await getAccounts(req);
+      const accounts: string[] = await getAccounts(req, false);
```
rekmarks, line 385: We should add a comment explaining what the `false` is doing here, I think.

rekmarks (changes requested):

src/wallet.ts
```diff
@@ -379,7 +382,7 @@ export function createWalletMiddleware({
   ): Promise<string> {
     if (typeof address === 'string' && address.length > 0) {
       // ensure address is included in provided accounts
```
shanejonas, line 384:
```suggestion
      // ensure address is included in provided accounts
      // `false` is passed to `getAccounts` to allow an unauthorized error to be thrown 
```

src/wallet.ts
rekmarks: A comment was added, but we'll still get the misleading error on L395 below if the address is valid and it's not contained in the array returned from `getAccounts`.

src/wallet.ts
```diff
@@ -21,7 +21,10 @@ interface TypedMessageParams extends MessageParams {
   version: string;
 }
 interface WalletMiddlewareOptions {
-  getAccounts: (req: JsonRpcRequest<unknown>) => Promise<string[]>;
+  getAccounts: (
+    req: JsonRpcRequest<unknown>,
+    suppressUnauthorized?: boolean,
```
mcmire, line 26: Maybe instead of a positional argument this should be in an object? `{ suppressUnauthorized = false }: { suppressUnauthorized?: boolean }`. That way when calling `getAccounts` with `false` it's a little more clear what the `false` is doing. Also see https://github.com/MetaMask/metamask-extension/pull/14126#discussion_r853335734

test/wallet.test.ts
```diff
@@ -156,6 +179,32 @@ describe('wallet', () => {
...
+      };
+
+      const payload = { method: 'eth_signTransaction', params: [txParams] };
+      try {
+        await pify(engine.handle).call(engine, payload);
+      } catch (e: any) {
+        expect(e.message).toEqual(
+          'The requested account and/or method has not been authorized by the user.',
+        );
+      }
```
mcmire, line 206: This test looks very, very similar to the happy path above. Are you sure this is throwing an error? P.S. it's better to write this as:

``` suggestion
      const promise = pify(engine.handle).call(engine, payload);
      await expect(promise).rejects.toThrow('The requested account and/or method has not been authorized by the user.')
```

test/wallet.test.ts
```diff
@@ -103,6 +103,29 @@ describe('wallet', () => {
...
+      };
+
+      const payload = { method: 'eth_sendTransaction', params: [txParams] };
+      try {
+        await pify(engine.handle).call(engine, payload);
+      } catch (e: any) {
+        expect(e.message).toEqual(
+          'The requested account and/or method has not been authorized by the user.',
+        );
+      }
```
mcmire, line 127: Similar to another comment, it's better to write this as the following to rule out a false positive:

``` suggestion
      const promise = pify(engine.handle).call(engine, payload);
      await expect(promise).rejects.toThrow('The requested account and/or method has not been authorized by the user.')
```

test/wallet.test.ts
```diff
@@ -223,6 +272,37 @@ describe('wallet', () => {
...
+        method: 'eth_signTypedData',
+        params: [message, testUnkownAddress],
+      };
+      try {
+        await pify(engine.handle).call(engine, payload);
+      } catch (e: any) {
+        expect(e.message).toEqual(
+          'The requested account and/or method has not been authorized by the user.',
+        );
+      }
```
mcmire, line 304: Similar to another comment, it's better to write this as the following to rule out a false positive:

``` suggestion
      const promise = pify(engine.handle).call(engine, payload);
      await expect(promise).rejects.toThrow('The requested account and/or method has not been authorized by the user.')
```

test/wallet.test.ts
```diff
@@ -283,6 +363,34 @@ describe('wallet', () => {
...
+        params: [message, testUnkownAddress],
+      };
+
+      try {
+        await pify(engine.handle).call(engine, payload);
+      } catch (e: any) {
+        expect(e.message).toEqual(
+          'The requested account and/or method has not been authorized by the user.',
+        );
+      }
```
mcmire, line 392:
``` suggestion
      const promise = pify(engine.handle).call(engine, payload);
      await expect(promise).rejects.toThrow('The requested account and/or method has not been authorized by the user.')
```

src/wallet.ts
mcmire: Oh I see, Erik already commented on this. Well, up to you, then.

mcmire (approved):
Seems kosher to me.

PR #118, opened 2022-04-06
Changed tests to jest + ts

shanejonas:
This changes the test to be `jest` and `ts`. We need this to better test my PR that adds an unauthorized error into the mix: https://github.com/MetaMask/eth-json-rpc-middleware/pull/116


test/wallet.test.ts
```diff
@@ -0,0 +1,346 @@
...
+  TransactionParams,
+  MessageParams,
+} from '../src';
+
+// eslint-disable-next-line
+const EthQuery = require('ethjs-query');
+// eslint-disable-next-line
+const pify = require('pify');
+
+// jest.useFakeTimers();
```
shanejonas, line 17:
```suggestion
```

test/wallet.test.ts
```diff
@@ -0,0 +1,345 @@
...
+  createWalletMiddleware,
+  TransactionParams,
+  MessageParams,
+} from '../src';
+
+// eslint-disable-next-line
+const EthQuery = require('ethjs-query');
+// eslint-disable-next-line
+const pify = require('pify');
+
```
shanejonas, line 16:
```suggestion
```

test/wallet.test.ts
```diff
@@ -0,0 +1,339 @@
+import { expect, describe, it } from '@jest/globals';
```
mcmire, line 1: The docs seem to indicate that importing the Jest globals is necessary if you plan on using ESM for Node, but since 1) we don't support current versions of Node yet and 2) we haven't configured TypeScript to spit out ESM, it doesn't seem like we need this yet. Was there a specific reason why you wanted to use this style?

test/wallet.test.ts
```diff
@@ -0,0 +1,339 @@
...
+
+import {
+  providerFromEngine,
+  createWalletMiddleware,
+  TransactionParams,
+  MessageParams,
+} from '../src';
+
+// eslint-disable-next-line
+const EthQuery = require('ethjs-query');
```
mcmire, line 12: Do imports not work for these dependencies?

jest.config.js
```diff
@@ -0,0 +1,8 @@
+// Sync object
```
mcmire, line 1: I've updated the module template with `jest.config.ts` — thoughts on using this? (You'll need `ts-node` to support this.)

shanejonas:
note: still need to convert the rest of these tests to ts:

image

test/block-ref.test.ts
```diff
@@ -0,0 +1,63 @@
...
+        providerReqs.push(_req);
+        cb(
+          undefined as any,
+          { id: _req.id, result: '0x0', jsonrpc: '2.0' } as any,
+        );
+      });
+
+    engine.push(createBlockRefMiddleware({ blockTracker, provider }));
+
+    // let hitCount = 0;
```
shanejonas, line 42:
```suggestion
```

test/block-ref.test.ts
```diff
@@ -0,0 +1,62 @@
...
+    });
+
+    await pify(engine.handle).call(engine, {
+      jsonrpc: '2.0',
+      id: 1,
+      method: 'eth_getBalance',
+      params: [testAddresses[0], 'latest'],
+    });
+
+    // expect(hitCountMiddleware.getHits('eth_getBalance').length).toEqual(1);
```
shanejonas, line 57:
```suggestion
```

package.json
```diff
@@ -32,8 +32,8 @@
     "lint:misc": "prettier '**/*.json' '**/*.md' '!CHANGELOG.md' '**/*.yml' --ignore-path .gitignore",
     "prepublishOnly": "yarn setup && yarn build:clean && yarn lint && yarn test:nobuild",
     "setup": "yarn install && yarn allow-scripts",
-    "test": "yarn build && yarn test:nobuild",
-    "test:nobuild": "node test/index.js"
+    "test": "yarn test:nobuild",
```
mcmire, line 35: Can we collapse these into one task? Additionally according to the template there are usually two tasks, `test` and `test:watch`.

test/inflight-cache.test.ts
```diff
@@ -0,0 +1,51 @@
...
+import pify from 'pify';
+import { providerFromEngine, createInflightCacheMiddleware } from '../src';
+
+function createTestSetup() {
+  // raw data source
+  // create block tracker
+  // create higher level
+  const engine = new JsonRpcEngine();
+  const provider = providerFromEngine(engine);
+  return { engine, provider };
```
mcmire, line 11: `provider` doesn't seem to be getting used in these tests.

test/util/createHitTrackerMiddleware.ts
```diff
@@ -0,0 +1,13 @@
+export default function createHitTrackerMiddleware() {
+  const hitTracker: any = {};
+  const middleware = (req: any, _res: any, next: any, _end: any) => {
```
mcmire, line 3: Is it possible to not use `any` here?

test/block-cache.test.ts
```diff
@@ -0,0 +1,60 @@
...
+
+function createTestSetup() {
+  // raw data source
+  // create block tracker
+  // create higher level
+  const engine = new JsonRpcEngine();
+  const provider = providerFromEngine(engine);
+
+  const blockTracker = new PollingBlockTracker({
+    provider: provider as Provider,
```
mcmire, line 15: I may know the answer to this already, but why do we need to typecast `provider`?

test/block-cache.test.ts
```diff
@@ -0,0 +1,60 @@
...
+  return { engine, provider, blockTracker };
+}
+
+describe('block cache', () => {
+  it('getBalance + undefined blockTag', async () => {
+    const { engine, provider, blockTracker } = createTestSetup();
+    const spy = jest
+      .spyOn(provider, 'sendAsync')
+      .mockImplementation((req, cb) => {
+        cb(undefined as any, { id: req.id, result: '0x0' } as any);
```
mcmire, line 28: Why do we need to typecast these?

src/fetch.ts
```diff
@@ -18,7 +18,7 @@ const RETRIABLE_ERRORS: string[] = [
   'Failed to fetch',
 ];
 
-interface PayloadwithOrgin extends Payload {
+export interface PayloadwithOrgin extends Payload {
```
mcmire, line 21: This is misspelled — should we fix the spelling?

test/block-cache.test.ts
```diff
@@ -0,0 +1,60 @@
...
+  const blockTracker = new PollingBlockTracker({
+    provider: provider as Provider,
+    pollingInterval: 200,
+  });
+
+  return { engine, provider, blockTracker };
+}
+
+describe('block cache', () => {
+  it('getBalance + undefined blockTag', async () => {
```
mcmire, line 23: Is it possible to give more informative names for these tests?

test/block-cache.test.ts
```diff
@@ -0,0 +1,60 @@
...
+function createTestSetup() {
+  // raw data source
+  // create block tracker
+  // create higher level
+  const engine = new JsonRpcEngine();
+  const provider = providerFromEngine(engine);
+
+  const blockTracker = new PollingBlockTracker({
+    provider: provider as Provider,
+    pollingInterval: 200,
```
mcmire, line 16: Does the polling interval matter in this case?

test/wallet.test.ts
```diff
@@ -0,0 +1,339 @@
...
+      engine.push(createWalletMiddleware({ getAccounts }));
+      const coinbaseResult = await pify(engine.handle).call(engine, {
+        method: 'eth_coinbase',
+      });
+      expect(coinbaseResult.result).toEqual(testAddresses[0]);
+    });
+  });
+
+  describe('transactions', () => {
+    it('valid address for processTransaction', async () => {
```
mcmire, line 62: What do you think about rewording this to describe the behavior and not merely how the test is being set up? I think there are some other examples in this file.

test/wallet.test.ts
```diff
@@ -0,0 +1,339 @@
...
+      } catch (e: any) {
+        expect(e.message).toEqual(
+          'Invalid parameters: must provide an Ethereum address.',
+        );
+      }
+    });
+  });
+
+  describe('personalRecover', () => {
+    it('geth kumavis manual recover', async () => {
```
mcmire, line 289: Hmm well this one may be difficult to describe, what is this even doing exactly 😆

mcmire (commented):
Looks a lot better. I like how you've structured these tests.

One overall thing other than the comments below: Should we be placing test files in the same directory as implementation files? This is what I've seen in other repos and I think we do the same thing in the template repo.

test/block-cache.test.ts
shanejonas: the `PollingBlockTracker` has a different provider type that the `error` doesnt match:

image

I think we want to replace this old provider interface with an EIP1193 compatible one everywhere, hopefully soon.

mcmire: Ugh... right. I've also noticed that the signature for the callback that `sendAsync` takes is incorrect, as apparently it wants to take an error object at all times. I'll make a note of this, thanks.

src/utils/cache.ts
```diff
@@ -17,18 +17,21 @@ export type BlockCache = Record<string, Block>;
...
+  providerRes: JsonRpcResponse<any>,
+) => void;
 
 export interface SafeEventEmitterProvider extends SafeEventEmitter {
-  sendAsync: (
-    req: JsonRpcRequest<string[]>,
-    callback: SendAsyncCallBack,
+  sendAsync: <T, U>(
+    req: JsonRpcRequest<T>,
+    cb: (err: unknown, response: JsonRpcResponse<U>) => void,
```
mcmire, line 32: What about:

``` suggestion
    cb: SendAsyncCallBack<U>,
```

src/utils/cache.ts
```diff
@@ -17,18 +17,21 @@ export type BlockCache = Record<string, Block>;
 export type Cache = Record<number, BlockCache>;
 
 export type SendAsyncCallBack = (
-  err: Error,
-  providerRes: PendingJsonRpcResponse<Block>,
+  err: any,
+  providerRes: JsonRpcResponse<any>,
 ) => void;
```
mcmire, line 22: Should this be:

``` suggestion
export type SendAsyncCallBack<T> = (
  err: unknown,
  providerRes: JsonRpcResponse<T>,
) => void;
```

mcmire (approved):
LGTM!

PR #124, opened 2022-04-27
Fix prepublishOnly script

rekmarks:
The `prepublishOnly` script was referencing a package script that no longer exists and, judging by its name, would be unnecessary anyway.


mcmire (approved):
Ah, I think Shane kept a `test:nobuild` at first in #118 and then I told him to remove it and didn't catch this. This makes sense though.

PR #125, opened 2022-04-27
Move eth-block-tracker to dependencies

### LINKED ISSUE #123: `eth-block-tracker` should be listed in `dependencies`, not `devDependencies`
We are attempting to convert `eth-json-rpc-infura` to TypeScript and are seeing build issues because of this: https://github.com/MetaMask/eth-json-rpc-infura/runs/6198120008?check_suite_focus=true

mcmire:
We are attempting to convert `eth-json-rpc-infura`, which relies on this
package, to TypeScript, and TypeScript is reporting errors resolving the
definition file for this package due to the types for
`eth-block-tracker` not being properly available. Moving it to
`dependencies` addresses this.

---

Fixes #123.


Gudahtt (approved):
LGTM!

PR #127, opened 2022-04-27
Expose SafeEventEmitterProvider type

### LINKED ISSUE #126: Add export for SafeEventEmitterProvider
This way `eth-json-rpc-infura` can make use of it. See: https://github.com/MetaMask/eth-json-rpc-infura/pull/58#discussion_r860092163

mcmire:
Not all of the types within `./src/utils/cache.ts` seemed to be solely
related to caching. These have been extracted to a separate file, and
the main file has been updated to export `SafeEventEmitterProvider` from
this file.

---

Fixes #126.


Gudahtt:
Looks like this PR isn't in a reviewable state right now because `dist` has been committed. That was a good idea to enable testing this change, but I think you can revert it on this branch and still leave the other PR pointing at the commit.

mcmire:
@Gudahtt Okay, should be ready for review now.

jest.config.ts
```diff
@@ -30,9 +30,7 @@ const config: Config.InitialOptions = {
   coverageDirectory: 'coverage',
 
   // An array of regexp pattern strings used to skip coverage collection
-  // coveragePathIgnorePatterns: [
-  //   "/node_modules/"
-  // ],
+  coveragePathIgnorePatterns: ['\\/src\\/types\\.ts$'],
```
Gudahtt, line 33: The forward slashes do not need to be escaped
```suggestion
  coveragePathIgnorePatterns: ['/src/types\\.ts$'],
```

mcmire: Ah right, I suppose since we're not using `/.../` it doesn't matter.

mcmire: Changed in https://github.com/MetaMask/eth-json-rpc-infura/pull/58 as well.

Gudahtt (approved):
LGTM!

PR #130, opened 2022-05-13
Update eth-sig-util to 3.0.1

heldersepu:
This is to fix a warning in my project, I get the following while building:
```
warning Pattern 
["ethereumjs-abi@git+https://github.com/ethereumjs/ethereumjs-abi.git"] 
is trying to unpack in the same destination 
"/home/heldersepu/.cache/yarn/v6/npm-ethereumjs-abi-0.6.8-ee3994657fa7a427238e6ba92a84d0b529bbcde0/node_modules/ethereumjs-abi" 
as pattern 
["ethereumjs-abi@^0.6.8"]. 
This could result in non-deterministic behavior, skipping.
```

That is because there are multiple versions of eth-sig-util used that in turn have different ethereumjs-abi...
This update to latest will fix it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "clone": "^2.1.1",
     "eth-block-tracker": "^5.0.1",
     "eth-rpc-errors": "^4.0.3",
-    "eth-sig-util": "^1.4.2",
+    "eth-sig-util": "^3.0.1",
     "json-rpc-engine": "^6.1.0",
     "json-stable-stringify": "^1.0.1",
     "node-fetch": "^2.6.7",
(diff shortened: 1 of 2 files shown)


mcmire:
Hi @heldersepu, thanks for this submission. `eth-sig-util` has been renamed to `@metamask/eth-sig-util` and the latest version is 4.0.1, so I think we ought to upgrade to that now. In addition, there have been breaking changes to the package which means that we need to slightly change how it's imported and used. Would you mind making these changes as well? You can read about them here: https://github.com/MetaMask/eth-sig-util/blob/main/CHANGELOG.md#401

PR #133, opened 2022-05-27
Upgrade to @metamask/eth-sig-util v4 and remove unused deps

kevinji:
This upgrades eth-sig-util to the latest version of @metamask/eth-sig-util. I also removed some unused dependencies that depended on this outdated eth-sig-util package and fixed an incorrect TypeScript type cast.

Closes #130.


kevinji:
@mcmire I noticed that you self-assigned this; is there anything I can do to help with the review?

Jandoria:
Hello, I'd love to have this PR merged, @mcmire I notice this has been opened for 3 months, is there anything I can do to help?

mcmire:
Hey, sorry for the wait on this, I just need to double-check this to ensure that these changes make sense. I haven't had time lately to do this, but I will try to make this a priority this week.

src/wallet.ts
```diff
@@ -310,16 +310,12 @@ export function createWalletMiddleware({
     req: JsonRpcRequest<unknown>,
     res: PendingJsonRpcResponse<unknown>,
   ): Promise<void> {
-    const message: string = (req.params as string)[0];
-    const signature: string = (req.params as string)[1];
-    const extraParams: Record<string, unknown> =
```
mcmire, line 315: I think there is a reason why we are accepting extra params, and I suspect we may need to keep this for backward compatibility. I would need to confirm on this.

kevinji: Gotcha, I removed it because @metamask/eth-sig-util only takes the [two parameters](https://github.com/MetaMask/eth-sig-util/blob/main/src/personal-sign.ts#L48), but it's possible that the API changed when the package got renamed.

Gudahtt: It looks like those extra parameters were never used in practice. I looked back through the history all the way to v1.x.

kevinji:
Just pushed a change to update the `yarn.lock` to fix the merge conflict.

kevinji:
Whoops I did not mean to upgrade all those dependencies. This new commit should do the trick.

src/wallet.ts
mcmire: Gotcha. I also checked too and this makes sense.

mcmire:
@kevinji If you have not already done so, would you mind running `yarn setup` instead of `yarn install`? If there are any errors, can you copy and paste those here?

kevinji:
@mcmire I did not run into any errors locally, just the following warnings:
```
[4/5] 🔗  Linking dependencies...
warning " > @metamask/eslint-config@8.0.0" has incorrect peer dependency "eslint@^7.23.0".
warning " > @metamask/eslint-config-nodejs@8.0.0" has incorrect peer dependency "eslint@^7.23.0".
warning " > @metamask/eslint-config-typescript@8.0.0" has incorrect peer dependency "eslint@^7.23.0".
[5/5] 🔨  Building fresh packages...
warning Ignored scripts due to flag.
```
Running on macOS if that matters.

mcmire (approved):

mcmire:
@kevinji Thanks!

It looks like the Socket checks are invalid. I checked (by looking at `yarn.lock` and running `yarn why`) and it looks like both `bufferutil` and `utf-8-validate` were removed entirely as dependencies.

mcmire:
@kevinji Thanks for the PR!

Issue #142, opened 2022-09-07
[block-ref] Duplicate request sent when `latest` or `undefined` block number given


Gudahtt:
When a request that accepts a block number parameter is given nothing or `latest` as the block number, the `block-ref` middleware is meant to replace that request with a new one that has a defined block number set that matches the latest block number the application is currently aware of (as informed by our block tracker).

Instead, currently what happens is that a duplicate request is sent with the defined block number, but the original request still goes through with a block number of `latest`. The response of the original request will overwrite the first one as well.

We should not be making duplicate requests, and we should be preserving the behaviour of replacing `latest` with our latest known block number.

Gudahtt:
Looks like this bug was introduced [here](https://github.com/MetaMask/eth-json-rpc-middleware/commit/76d676566b0b2be75bc0e952789f4442990a0daf#diff-00d71f9a810b6015201c9aae4ec2d2a5428cc7e2d7745e4267bc36ad58d86bf6R42) in v7.0.0

PR #147, opened 2022-09-26
Fix bugs in `retryOnEmpty` middleware

### LINKED ISSUE #139: `retryOnEmpty` incorrectly checking the response from the child request
In https://github.com/MetaMask/eth-json-rpc-middleware/commit/3eb4f1a8580e2190d0dddddee4a80daa182065c6#diff-9043aff30876c1d6c43ddee1949bf50ced00c070e176a175ce46e0b418a9d61cR77, the retry-on-empty middleware was converted to TypeScript. In the [original (JavaScript) version](https://github.com/MetaMask/eth-json-rpc-middleware/blob/ce8a476947d0cc519b03fcb5ae4cd5c5e389f88f/src/retryOnEmpty.js#L62) of this middleware, it seems that we were checking whether the `result` in the response was empty. Now we are checking whether the _whole response_ is empty. This seems like a bug, and it makes this middleware effectively useless in the extension (because thanks to the Infura middleware as exported from `createInfuraMiddleware` in `eth-json-rpc-infura`, the result of `provider.sendAsync` will _always_ be a valid JSON-RPC response object and never null or undefined).

mcmire:
When this package was converted to TypeScript, two bugs were introduced
in the `retryOnEmpty` middleware.

* The goal of this middleware is to retry requests to the network that
return responses whose `result` property is empty in some way, but after
the TypeScript conversion, we were checking whether the whole *response*
object was empty rather than just the `result`. This never happens in
practice, so it ended up making the middleware ineffective.
* Additionally, after retrying the request, the middleware fell through
to the next middleware, which in practice is the middleware that makes a
request to Infura. This meant that later middleware overwrote the
results of the `retryOnEmpty` middleware, also making it ineffective.

This commit restores the functionality prior to the TypeScript
conversion and adds tests to confirm.

---

Fixes #139.

### DIFF
--- a/test/retryOnEmpty.test.ts
+++ b/test/retryOnEmpty.test.ts
@@ -493,9 +493,9 @@ function requestMatches(
   requestMatcher: Partial<JsonRpcRequest<unknown>>,
   request: JsonRpcRequest<unknown>,
 ): boolean {
-  return (
-    Object.keys(requestMatcher) as unknown as (keyof typeof requestMatcher)[]
-  ).every((key) => isDeepStrictEqual(requestMatcher[key], request[key]));
+  return (Object.keys(requestMatcher) as (keyof typeof requestMatcher)[]).every(
+    (key) => isDeepStrictEqual(requestMatcher[key], request[key]),
+  );
 }
 
 /**
--- a/test/retryOnEmpty.test.ts
+++ b/test/retryOnEmpty.test.ts
@@ -1,0 +1,802 @@
+import { isDeepStrictEqual } from 'util';
+import clone from 'clone';
+import { PollingBlockTracker, Provider } from 'eth-block-tracker';
+import {
+  JsonRpcEngine,
+  JsonRpcRequest,
+  JsonRpcResponse,
+} from 'json-rpc-engine';
+import {
+  providerFromEngine,
+  createRetryOnEmptyMiddleware,
+  SafeEventEmitterProvider,
+} from '../src';
+
+/**
+ * Objects used in each test.
+ *
+ * @property engine - The engine that holds the middleware stack, including the
+ * one being tested.
+ * @property provider - The provider that is used to make requests against
+ * (which the middleware being tested will react to).
+ * @property blockTracker - The block tracker which is used inside of the
+ * middleware being tested.
+ */
+interface Setup {
+  engine: JsonRpcEngine;
+  provider: SafeEventEmitterProvider;
+  blockTracker: PollingBlockTracker;
+}
+
+/**
+ * Options supported by `withTestSetup`.
+ *
+ * @property withFinalMiddleware - Whether or not the middleware stack should be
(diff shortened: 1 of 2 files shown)


test/retryOnEmpty.test.ts
```diff
@@ -0,0 +1,668 @@
...
+
+      if (res.result === undefined) {
+        res.result = 'default response';
+      }
+
+      end();
+    });
+  }
+
+  return await fn(setup);
```
mcmire, line 436: For now there is no reason to have this function take a function — it could just return `setup`. However, this package relies on an older version of `eth-block-tracker`, so once we upgrade that package we can call `blockTracker.destroy` here, and that should make this follow the same sort of pattern that we've followed in other projects.

test/retryOnEmpty.test.ts
```diff
@@ -0,0 +1,668 @@
...
+
+            expect(await promiseForResponse).toMatchObject({
+              error: expect.objectContaining({
+                message: 'RetryOnEmptyMiddleware - retries exhausted',
+              }),
+            });
+          });
+        });
+
+        it('falls through to the next middleware after making a request directly through the provider', async () => {
```
Gudahtt, line 207: This behavior is not intended. It's caused by another bug that was introduced around the same time - the same type of bug as #142 (and introduced in the same PR)

The final `next()` after a successful child request should not be there, because calling that results in the response from the child request being overwritten. It makes this whole middleware pointless.

mcmire: That's what I figured, but I wasn't sure 😵

test/retryOnEmpty.test.ts
```diff
@@ -0,0 +1,668 @@
...
+ * @param request - A real request object.
+ * @returns True or false depending on whether the partial request object "fits
+ * inside" the real request object.
+ */
+function requestMatches(
+  requestMatcher: Partial<JsonRpcRequest<unknown>>,
+  request: JsonRpcRequest<unknown>,
+): boolean {
+  return (
+    Object.keys(requestMatcher) as unknown as (keyof typeof requestMatcher)[]
```
Gudahtt, line 373: Nit: I don't think the `as unknown` part here is required:
```suggestion
    Object.keys(requestMatcher)  as (keyof typeof requestMatcher)[]
```

test/retryOnEmpty.test.ts
mcmire: I've fixed this bug in this PR as well and updated the PR description to match.

test/retryOnEmpty.test.ts
mcmire: Hmm, weird. Not sure why I thought this was necessary. Removed in [e4f373b](https://github.com/MetaMask/eth-json-rpc-middleware/pull/147/commits/e4f373b0462f4b653714ba58c358574a85048f61).

PR #150, opened 2022-09-29
Bump @metamask/eth-sig-util to 5.0.0

mcmire:
It seems that we only use `recoverPersonalSignature` from `eth-sig-util`, so the recent changes around `@ethereumjs/util` do not seem to affect this library.


Gudahtt (approved):
LGTM!

PR #151, opened 2022-09-30
Update block-ref to not make duplicate request

### LINKED ISSUE #142: [block-ref] Duplicate request sent when `latest` or `undefined` block number given

mcmire:
The goal of the `block-ref` middleware is to look for instances in request parameters where "latest" is being used as a block parameter and replace it with the actual latest block number (as reported by the block tracker). The middleware will then make a child request through the provider using the resolved parameters and copy the child response onto the final response object.

That is fine; however, this middleware then calls the next middleware in the stack, which in practice will make another request, and that will end up overwriting the child response set in this middleware. That makes this middleware ineffective.

So, this commit removes the `next()` call from the `block-ref` middleware so that when it takes effect it stops the chain and prevents an extra request.

In addition, this fills in missing tests for the middleware by reusing some test helpers from a previous PR.

---

Fixes #142.

### DIFF
--- a/src/block-ref.test.ts
+++ b/src/block-ref.test.ts
@@ -1,59 +1,392 @@
+import { isDeepStrictEqual } from 'util';
 import { PollingBlockTracker, Provider } from 'eth-block-tracker';
-import { JsonRpcEngine } from 'json-rpc-engine';
-import pify from 'pify';
-import { providerFromEngine, createBlockRefMiddleware } from '.';
+import {
+  JsonRpcEngine,
+  JsonRpcMiddleware,
+  JsonRpcRequest,
+  JsonRpcResponse,
+} from 'json-rpc-engine';
+import clone from 'clone';
+import {
+  SafeEventEmitterProvider,
+  providerFromEngine,
+  createBlockRefMiddleware,
+} from '.';
 
-const testAddresses = [
-  '0xbe93f9bacbcffc8ee6663f2647917ed7a20a57bb',
-  '0x1234362ef32bcd26d3dd18ca749378213625ba0b',
-];
+/**
+ * Objects used in each test.
+ *
+ * @property engine - The engine that holds the middleware stack, including the
+ * one being tested.
+ * @property provider - The provider that is used to make requests against
+ * (which the middleware being tested will react to).
+ * @property blockTracker - The block tracker which is used inside of the
+ * middleware being tested.
+ */
+interface Setup {
+  engine: JsonRpcEngine;
+  provider: SafeEventEmitterProvider;
+  blockTracker: PollingBlockTracker;
+}
 
-function createTestSetup() {
-  // raw data source
-  // create block tracker
-  // create higher level
+/**
+ * The function that `withTestSetup` is expected to take and will call once the
+ * setup objects are created.
+ */
+type WithTestSetupCallback<T> = (setup: Setup) => Promise<T>;
+
+/**
+ * An object that can be used to assign a canned response to a request (or an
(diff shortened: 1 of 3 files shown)


src/block-ref.ts
```diff
@@ -31,38 +25,31 @@ export function createBlockRefMiddleware({
...
-    // perform child request
-    log('Performing another request %o', childRequest);
-    const childRes: PendingJsonRpcResponse<Block> = await pify(
-      (provider as SafeEventEmitterProvider).sendAsync,
-    ).call(provider, childRequest);
-    // copy child response onto original response
-    res.result = childRes.result;
-    res.error = childRes.error;
+
+    req.params[blockRefIndex] = latestBlock;
```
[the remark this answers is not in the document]
mcmire: Sigh... that's very strange

Okay, I've reverted the behavior and fixed up the tests accordingly: [737f54e](https://github.com/MetaMask/eth-json-rpc-middleware/pull/151/commits/737f54eaa094a4c3cdbd709002835beaa3d943d1)

src/block-ref.test.ts
```diff
@@ -1,59 +1,620 @@
...
+      { blockParamIndex: 2, methods: ['eth_getStorageAt'] },
+    ] as const
+  ).forEach(({ blockParamIndex, methods }) => {
+    methods.forEach((method: string) => {
+      describe(`when the RPC method is ${method}`, () => {
+        describe('if the block param is "latest"', () => {
+          it('makes a direct request through the provider, replacing the block param with the latest block number', async () => {
+            await withTestSetup(
+              {
+                configureMiddleware: ({ provider, blockTracker }) => {
```
mcmire, line 105: This is a slightly different style compared with the `retryOnEmpty` tests I wrote earlier. I wanted to make it so that:

* It's obvious what the middleware-under-test is. Previously this was automatically added in a helper function, now it happens within each test explicitly. Yet, it's drawn out so it's easier to see.
* If you don't need to supply a final middleware (because you're not testing that the middleware-under-test is calling the next middleware), you don't need to supply an extra option like `withFinalMiddleware: false`. At the same time, a default middleware is still pushed onto the engine. The purpose of this default middleware is to serve as a reminder for the developer in case the middleware-under-test attempts to call the next middleware and that was unexpected.
* If you do need to supply a final middleware because you want to test that it gets called, then you can still do that easily.

src/block-ref.test.ts
```diff
@@ -1,59 +1,620 @@
...
+ * @template T - The type that the function will return, minus the promise
+ * wrapper.
+ * @param options - Options.
+ * @param options.configureMiddleware - A function that is called to add the
+ * middleware-under-test, along with any other necessary middleware, to the
+ * engine.
+ * @param callback - A function.
+ * @returns Whatever the function returns.
+ */
+async function withTestSetup<T>(
```
mcmire, line 412: A lot of the helper functions here were copied from the tests for `retryOnEmpty`. I plan on introducing another PR after this one which will extract these helpers from the two files.

Gudahtt (approved):
LGTM!

PR #156, opened 2022-10-05
Update new tests to destroy block tracker

mcmire:
Certain middleware use a polling block tracker to retrieve the latest block. This type of block tracker has the ability to make a quick request in addition to polling, but doing so requires starting the poll loop and then immediately stopping it. This means that it is possible for the block tracker to be in a running state when a test that uses it ends. To avoid this, we need to make sure we completely shut down the block tracker at the end of each test. The ability to do this was added in a recent version of `eth-block-tracker`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "@metamask/utils": "^3.0.3",
     "btoa": "^1.2.1",
     "clone": "^2.1.1",
-    "eth-block-tracker": "^5.0.1",
+    "eth-block-tracker": "^6.0.0",
     "eth-rpc-errors": "^4.0.3",
     "json-rpc-engine": "^6.1.0",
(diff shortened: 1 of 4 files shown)


mcmire:
@Gudahtt Hmm, these changes don't as useful as I thought. After running the test suite on this branch, I'm seeing the "Jest did not exit one second after the test run has completed" message from Jest, which I feel like these changes were intended to remove. I've played around with this a bit, and it seems that destroying the PollingBlockTracker doesn't kill the last `setTimeout` call it makes when it advances to the next iteration of the polling loop. We can "fix" this by instantiating PollingBlockTracker with `keepEventLoopActive: false`, but that papers over the problem in my opinion. So I'm going to keep this open until `eth-block-tracker` is properly fixed.

Issue #176, opened 2023-01-30
Fix middleware types


Gudahtt:
The middleware types include a number of errors that are making them difficult to use in TypeScript projects.

The types suggest that all requests would have a list of string parameters, but this is not guaranteed or even expected. Minimally we'll need to update them to expect `unknown` parameters.

PR #177, opened 2023-01-30
Refactor cache helper methods

Gudahtt:
Refactor cache helper methods to no longer use the `Payload` type to access the request method. Instead they now take the method directly.

This relates to #176, in that it reduces the number of places where we are using this `Payload` type that we need to eliminate for #176.

### DIFF
--- a/src/utils/cache.test.ts
+++ b/src/utils/cache.test.ts
@@ -1,0 +1,145 @@
+import { blockTagParamIndex, cacheTypeForMethod } from './cache';
+
+const knownMethods = [
+  'web3_clientVersion',
+  'web3_sha3',
+  'eth_protocolVersion',
+  'eth_getBlockTransactionCountByHash',
+  'eth_getUncleCountByBlockHash',
+  'eth_getCode',
+  'eth_getBlockByHash',
+  'eth_getTransactionByHash',
+  'eth_getTransactionByBlockHashAndIndex',
+  'eth_getTransactionReceipt',
+  'eth_getUncleByBlockHashAndIndex',
+  'eth_getCompilers',
+  'eth_compileLLL',
+  'eth_compileSolidity',
+  'eth_compileSerpent',
+  'shh_version',
+  'test_permaCache',
+  'eth_getBlockByNumber',
+  'eth_getBlockTransactionCountByNumber',
+  'eth_getUncleCountByBlockNumber',
+  'eth_getTransactionByBlockNumberAndIndex',
+  'eth_getUncleByBlockNumberAndIndex',
+  'test_forkCache',
+  'eth_gasPrice',
+  'eth_blockNumber',
+  'eth_getBalance',
+  'eth_getStorageAt',
+  'eth_getTransactionCount',
+  'eth_call',
+  'eth_estimateGas',
+  'eth_getFilterLogs',
+  'eth_getLogs',
+  'test_blockCache',
+];
+
+describe('cache utils', () => {
+  describe('blockTagParamIndex', () => {
+    it(`should return expected block index for each known method`, () => {
+      const blockTagIndexes = knownMethods.reduce((indexes, method) => {
+        indexes[method] = blockTagParamIndex(method);
+        return indexes;
+      }, {} as Record<string, number | undefined>);
+
+      expect(blockTagIndexes).toMatchInlineSnapshot(`
+        Object {
+          "eth_blockNumber": undefined,
+          "eth_call": 1,
+          "eth_compileLLL": undefined,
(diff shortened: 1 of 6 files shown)


src/utils/cache.test.ts
```diff
@@ -0,0 +1,157 @@
...
+  'eth_getBlockByHash',
+  'eth_getTransactionByHash',
+  'eth_getTransactionByBlockHashAndIndex',
+  'eth_getTransactionReceipt',
+  'eth_getUncleByBlockHashAndIndex',
+  'eth_getCompilers',
+  'eth_compileLLL',
+  'eth_compileSolidity',
+  'eth_compileSerpent',
+  'shh_version',
```
shanejonas, line 19: `shh_version` doesn't exist anymore.

tried calling infura with it:

image

There are more too, `eth_compileLLL` and the other `compileX` dont exist anymore either.

we should probably get our method lists for eth rpc from the spec: https://github.com/ethereum/execution-apis

Gudahtt: By "known methods" here, I meant known by the middleware rather than us. This list is taken directly from the code under test.

Definitely lots of updates needed to this middleware but I'd rather not take that work on right now. Not until after we finish this multichain work.

src/utils/cache.ts
```diff
@@ -8,21 +8,27 @@ export function cacheIdentifierForPayload(
...
 
-export function canCache(payload: Payload): boolean {
-  return cacheTypeForPayload(payload) !== 'never';
+/**
+ * Return whether a method can be cached or not.
+ *
+ * @param method - The method to check.
+ * @returns Whether the method can be cached.
+ */
+export function canCache(method?: string): boolean {
```
mcmire, line 23: Why is this argument optional? Same for the other functions here.

If this can be undefined because `Payload.method` itself is optional, then I'm wondering whether this would make more sense conceptually:

``` suggestion
export function canCache(method: string | undefined): boolean {
```

That way `canCache(undefined)` would be a valid way to call this function, but `canCache()` would not.

src/utils/cache.test.ts
```diff
@@ -0,0 +1,157 @@
...
+  describe('cacheTypeForMethod', () => {
+    it(`should return expected cache type for each known method`, () => {
+      const cacheTypes = knownMethods.reduce((types, method) => {
+        types[method] = cacheTypeForMethod(method);
+        return types;
+      }, {} as Record<string, string>);
+
+      expect(cacheTypes).toMatchInlineSnapshot(`
+        Object {
+          "eth_blockNumber": "block",
```
mcmire, line 114: I like this a lot.

src/utils/cache.ts
```diff
@@ -70,8 +83,14 @@ export function blockTagParamIndex(payload: Payload): number | undefined {
...
 
-export function cacheTypeForPayload(payload: Payload): string {
-  switch (payload.method) {
+/**
+ * Return the cache type used for the given method.
+ *
+ * @param method - A JSON-RPC method.
+ * @returns The cache type to use for that method.
+ */
+export function cacheTypeForMethod(method?: string): string {
```
mcmire, line 92: Really curious why we decided to have this return `string` rather than `"perma" | "fork" | "block" | "never"`. Sigh...

mcmire (commented):
One minor comment, but the rest looks good.

src/utils/cache.ts
Gudahtt: Yeah it's because of the Payload type, as you guessed. I hadn't bothered with this because I'm going to make this mandatory again once we get rid of Payload. I've almost gotten rid of it already in the latest PR.

src/utils/cache.ts
Gudahtt: Yeah, missed opportunity. This is something I wanted to do as well - add an enum for this.

src/utils/cache.ts
Gudahtt: The soonest I can remove this is in #182. I've just updated that PR with a commit to do just that: https://github.com/MetaMask/eth-json-rpc-middleware/pull/182/commits/bbd4931b6f7889a6a72903c0c7c1db4186374390

src/utils/cache.ts
Gudahtt: Done here: https://github.com/MetaMask/eth-json-rpc-middleware/pull/184

src/utils/cache.ts
mcmire: Gotcha, makes sense.

mcmire (approved):
Looks good!

PR #178, opened 2023-01-30
Fix types for middleware supporting `skipCache` property

Gudahtt:
Two middleware functions accept an additional request property called `skipCache`. Previously the return types for these middleware did not include this property. It was supported internally only by casting.

The return type of this middleware has been updated to reflect that this property is accepted. It has also been made optional. It was optional in practice already, this just wasn't reflected in the internal type used.

This changes the return type of the `block-cache` and `inflight-cache` middleware, but the new types should be fully compatible with the old types.


Gudahtt:
This depends upon #177

src/types.ts
```diff
@@ -1,12 +1,23 @@
...
-export interface JsonRpcRequestToCache extends JsonRpcRequest<string[]> {
-  skipCache: boolean;
+export interface JsonRpcRequestToCache<T> extends JsonRpcRequest<T> {
+  skipCache?: boolean;
 }
 
+export type JsonRpcCacheMiddleware<T, U> = JsonRpcMiddleware<T, U> extends (
+  req: JsonRpcRequest<T>,
+  ...args: infer X
+) => infer Y
```
mcmire, line 17: According to JsonRpcMiddleware, [the middleware function always returns void](https://github.com/MetaMask/json-rpc-engine/blob/a74885c620d06af96197541994cb06bc5018b56d/src/JsonRpcEngine.ts#L34-L42), so do we have to infer `Y`?

Besides this, I'm not sure I quite understand what this is doing. How would we end up getting `never` here?

src/block-cache.ts
```diff
@@ -156,80 +157,82 @@ export function createBlockCacheMiddleware({
...
-    }
-    // check type and matching strategy
-    const type: string = cacheTypeForMethod(req.method);
-    const strategy: BlockCacheStrategy = strategies[type];
-    // If there's no strategy in place, pass it down the chain.
-    if (!strategy) {
-      return next();
-    }
+  return createAsyncMiddleware(
+    async (req: JsonRpcRequestToCache<string[]>, res, next) => {
```
Gudahtt, line 161: I'd recommend using "hide whitespace" when viewing this diff; most of these changes are just indentation resulting from the change on this line

src/types.ts
Gudahtt: This is a way of exchanging the type of the first parameter without touching the rest. That's the only type difference with this middleware: it accepts a non-standard request property.

To accomplish this, I'm inferring the remainder of the parameters and the return value.

The `never` case never happens, as the name might imply. It's just that a conditional type is the only way I know how infer part of the structure of a type: by taking the original type and asking if it `extends` another type that includes the inferred pieces you want to build with, and if so returning a new type using those pieces.

The `never` case is to ensure any assumptions embedded here remain true, otherwise this will cause a type error. For example this ensures that this type remains valid even if the return type _does_ change. But it will throw an error if the request field we're replacing here changes.

Gudahtt: For more on this: https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#inferring-within-conditional-types

That's the only place I could find in the official docs on doing this, but I came across some more detailed guides elsewhere. Can't recall offhand where.

mcmire: Ah, gotcha. I missed the difference between `req: JsonRpcRequest<T>` and `JsonRpcRequestToCache<T>`. That makes more sense!

mcmire (approved):
I really like how well this cleans things up. I'll have to remember this solution for the future.

PR #179, opened 2023-01-30
Update repository to match latest module template

Gudahtt:
The GitHub actions have been updated to match the latest module template. Notable changes are that the checks now run in parallel, they include Node.js versions v18 and v19, and the release automation now includes the npm publish step.

The gitattributes file was also updated to include entries for Yarn v3 that were missed, and the Yarn config was updated to disable telemetry.


.github/workflows/publish-release.yml
```diff
@@ -1,29 +1,92 @@
...
+        with:
+          # This `NPM_TOKEN` needs to be manually set per-repository.
+          # Look in the repository settings under "Environments", and set this token in the `npm-publish` environment.
+          npm-token: ${{ secrets.NPM_TOKEN }}
+        env:
+          SKIP_PREPACK: true
+
+  get-release-version:
+    runs-on: ubuntu-latest
+    needs: publish-npm
```
Gudahtt, line 83: I setup this environment, but I can't generate the secret right now because of the pending rename. We've moved to using granular access tokens (restricted to a single package or set of packages), and I can't generate it for a package that doesn't exist yet.

I'll let this fall through to a dry run on the next release, and publish it manually. Then I should be able to generate and add the token for next time.

mcmire (approved):

.github/workflows/publish-release.yml
Gudahtt: This has now been setup

PR #181, opened 2023-01-30
Rename `blockTagForPayload` to `blockTagForRequest`

Gudahtt:
The utility method `blockTagForPayload` has been renamed to `blockTagForRequest`, and updated to accept a JSON-RPC request rather than a `Payload`. Documentation and tests have been added as well.

Relates to #176, which is partially caused by this `Payload` type that we're removing one usage of here.

### DIFF
--- a/src/utils/cache.ts
+++ b/src/utils/cache.ts
@@ -53,9 +53,7 @@ export function canCache(method?: string): boolean {
  * @param request - The JSON-RPC request.
  * @returns The block parameter in the given request, or `undefined` if none was found.
  */
-export function blockTagForRequest(
-  request: JsonRpcRequest<unknown>,
-): string | undefined {
+export function blockTagForRequest(request: JsonRpcRequest<unknown>): unknown {
   if (!request.params) {
     return undefined;
   }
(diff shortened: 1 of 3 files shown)


Gudahtt:
This depends upon #178

src/utils/cache.ts
```diff
@@ -24,18 +25,30 @@ export function canCache(method?: string): boolean {
...
-  if (!payload.params) {
+/**
+ * Return the block parameter for the given request, if it has one.
+ *
+ * @param request - The JSON-RPC request.
+ * @returns The block parameter in the given request, or `undefined` if none was found.
+ */
+export function blockTagForRequest(
+  request: JsonRpcRequest<unknown>,
+): string | undefined {
```
mcmire, line 36: Hmm. How does TypeScript know that the block param will be a string if the params type parameter is `unknown`?

I'm wondering if it makes more sense to do:

``` suggestion
export function blockTagForRequest<T>(
  request: JsonRpcRequest<unknown>,
): T | undefined {
```

or even:

``` suggestion
export function blockTagForRequest(
  request: JsonRpcRequest<unknown>,
): unknown | undefined {
```

Gudahtt: .... huh. That's a great point. Not sure why it let me do that.

Gudahtt: Wow, apparently TypeScript will let me set this left type to anything! No errors. That seems like a bug.

mcmire (approved):
Looks good!

PR #182, opened 2023-01-31
Refactor function to get cache identifier

Gudahtt:
The function `cacheIdentifierForPayload` has been renamed to `cacheIdentifierForRequest`, and refactored to accept a JSON-RPC request rather than the `Payload` type.

The function `paramsWithoutBlockTag` has been updated to accept a JSON- RPC request rather than a `Payload` as well. I noticed that this function wasn't being used anywhere, so the `export` was removed as well.

The `block-cache` and `inflight-cache` middleware required some internal updates to accomodate these type changes. No functional changes or external-facing type changes have been made.

Documentation and tests have been added. This brings test coverage for `cache.ts` up to 100%.

Relates to #176 in that it helps unblock the necessary changes.

PR #183, opened 2023-01-31
Use unknown JSON-RPC requests and responses

### LINKED ISSUE #176: Fix middleware types

Gudahtt:
The types of all middleware have been updated to remove assumptions about the parameter and response types. Previously the parameters were all assumed to be `string[]` and the response results were assumed to be `Block`, but neither of those are true in practice.

Fixes #176

### DIFF
--- a/src/block-ref-rewrite.ts
+++ b/src/block-ref-rewrite.ts
@@ -1,7 +1,6 @@
 import { PollingBlockTracker } from 'eth-block-tracker';
 import { createAsyncMiddleware, JsonRpcMiddleware } from 'json-rpc-engine';
 import { blockTagParamIndex } from './utils/cache';
-import type { Block } from './types';
 
 interface BlockRefRewriteMiddlewareOptions {
   blockTracker?: PollingBlockTracker;
@@ -9,7 +8,7 @@ interface BlockRefRewriteMiddlewareOptions {
 
 export function createBlockRefRewriteMiddleware({
   blockTracker,
-}: BlockRefRewriteMiddlewareOptions = {}): JsonRpcMiddleware<string[], Block> {
+}: BlockRefRewriteMiddlewareOptions = {}): JsonRpcMiddleware<unknown, unknown> {
   if (!blockTracker) {
     throw Error(
       'BlockRefRewriteMiddleware - mandatory "blockTracker" option is missing.',
@@ -23,7 +22,9 @@ export function createBlockRefRewriteMiddleware({
       return next();
     }
     // skip if not "latest"
-    let blockRef: string | undefined = req.params?.[blockRefIndex];
+    let blockRef: string | undefined = Array.isArray(req.params)
+      ? req.params[blockRefIndex]
+      : undefined;
     // omitted blockRef implies "latest"
     if (blockRef === undefined) {
       blockRef = 'latest';
@@ -34,7 +35,7 @@ export function createBlockRefRewriteMiddleware({
     }
     // rewrite blockRef to block-tracker's block number
     const latestBlockNumber = await blockTracker.getLatestBlock();
-    if (req.params) {
+    if (req.params && Array.isArray(req.params)) {
       // eslint-disable-next-line require-atomic-updates
       req.params[blockRefIndex] = latestBlockNumber;
     }
--- a/src/block-ref-rewrite.ts
+++ b/src/block-ref-rewrite.ts
@@ -35,7 +35,7 @@ export function createBlockRefRewriteMiddleware({
     }
     // rewrite blockRef to block-tracker's block number
     const latestBlockNumber = await blockTracker.getLatestBlock();
-    if (req.params && Array.isArray(req.params)) {
+    if (Array.isArray(req.params)) {
       // eslint-disable-next-line require-atomic-updates
       req.params[blockRefIndex] = latestBlockNumber;
     }
--- a/src/block-ref.ts
+++ b/src/block-ref.ts
@@ -58,7 +58,7 @@ export function createBlockRefMiddleware({
     // create child request with specific block-ref
     const childRequest = clone(req);
 
-    if (childRequest.params && Array.isArray(childRequest.params)) {
+    if (Array.isArray(childRequest.params)) {
       childRequest.params[blockRefIndex] = latestBlockNumber;
     }
 
--- a/src/block-ref.ts
+++ b/src/block-ref.ts
@@ -20,7 +20,7 @@ const log = createModuleLogger(projectLogger, 'block-ref');
 export function createBlockRefMiddleware({
   provider,
   blockTracker,
-}: BlockRefMiddlewareOptions = {}): JsonRpcMiddleware<string[], Block> {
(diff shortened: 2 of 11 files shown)


Gudahtt:
This depends upon #182

src/block-tracker-inspector.ts
```diff
@@ -13,37 +16,64 @@ interface BlockTrackerInspectorMiddlewareOptions {
...
 }
 
+function hasProperty<Property extends string>(
+  object: unknown,
+  property: Property,
+): object is Record<Property, unknown> {
+  return Object.hasOwnProperty.call(object, property);
+}
+
+function getResultBlockNumber(
```
Gudahtt, line 26: The type narrowing here was painful to work through. Not sure if there's an easier way to do this

src/block-tracker-inspector.ts
```diff
@@ -13,37 +16,64 @@ interface BlockTrackerInspectorMiddlewareOptions {
   blockTracker: PollingBlockTracker;
 }
 
+function hasProperty<Property extends string>(
+  object: unknown,
+  property: Property,
+): object is Record<Property, unknown> {
```
Gudahtt, line 22: This is similar to the one in our utils library: https://github.com/MetaMask/utils/blob/main/src/misc.ts#L93
Except here I used a type predicate to assert that the object had this key. The type narrowing in `getResultBlockNumber` doesn't work without it.

src/block-ref-rewrite.ts
```diff
@@ -34,7 +35,7 @@ export function createBlockRefRewriteMiddleware({
     }
     // rewrite blockRef to block-tracker's block number
     const latestBlockNumber = await blockTracker.getLatestBlock();
-    if (req.params) {
+    if (req.params && Array.isArray(req.params)) {
```
legobeat, line 38: Nit:
```suggestion
    if (Array.isArray(req.params)) {
```

src/block-ref.ts
```diff
@@ -56,7 +58,7 @@ export function createBlockRefMiddleware({
     // create child request with specific block-ref
     const childRequest = clone(req);
 
-    if (childRequest.params) {
+    if (childRequest.params && Array.isArray(childRequest.params)) {
```
legobeat, line 61:
```suggestion
    if (Array.isArray(childRequest.params)) {
```

src/block-tracker-inspector.ts
legobeat: I like the alternative suggestion, as it's a bit more semantic and descriptive for any users making their way to the types.

src/block-tracker-inspector.ts
```diff
@@ -13,37 +16,64 @@ interface BlockTrackerInspectorMiddlewareOptions {
...
+  const { result } = response;
+  if (
+    !result ||
+    typeof result !== 'object' ||
+    !hasProperty(result, 'blockNumber')
+  ) {
+    return undefined;
+  }
+
+  if (result.blockNumber && typeof result.blockNumber === 'string') {
```
legobeat, line 38:
```suggestion
  if (typeof result.blockNumber === 'string') {
```

Redundant since we already check for `result` just before?

Gudahtt: Good catch - it looks like this can be removed

src/block-tracker-inspector.ts
Gudahtt: Done here: 5b5f61854ca3af20b2e44da1cc4cde947506a61e

I realized that the constraint on `Property` was too restrictive in that it prevented `number` and `symbol` properties, so I moved the valid property types to a separate type and referenced that in both places.

src/block-ref-rewrite.ts
Gudahtt: Resolved here: 5ebedbf7a144f78a30749198d3441c0ae41afa48

src/block-tracker-inspector.ts
Gudahtt: Resolved here: 5ebedbf7a144f78a30749198d3441c0ae41afa48

src/block-ref.ts
Gudahtt: Resolved here: 5ebedbf7a144f78a30749198d3441c0ae41afa48

src/block-tracker-inspector.ts
mcmire: Hmm. It seems that when using this function, `object` would become `never` when the function returns false. I don't think you need to redeclare the type of the object entirely — I think you can use `&` to extend its definition. What about this?

```
type ValidPropertyType = string | number | symbol;

function hasProperty2<Object, Property extends ValidPropertyType>(
  object: Object,
  name: Property,
): object is Object & Record<Property, unknown> {
  return Object.hasOwnProperty.call(object, name);
}
```

See playground: https://www.typescriptlang.org/play?ts=4.9.4&ssl=1&ssc=1&pln=2&pc=1#code/C4TwDgpgBAaghgGwJYBMAKAnA9pDoAq40AvFAM7AZIB2A5lAD5TUCuAtgEYQaPkidYEAbgBQIgGYtqAY2BIs1KAAs4ZTDm6gAjAB51uUFAgAPYBGooysRKn2aQhSAD4AFCKhQsHAFYRZALigpAGtqLAB3agAad2Y4NghAuzwQGIBKQK9fWSgkKwAlPywMFD1sAxBeeGR0cvtHCCig6lCI6icoAG9YjAhgFgxFAHkfP2AAOhUyIcjk0HHpRAQXLLGm6niINNEAXzFJGTkFZVU5kAAmHRHs4Cazo1NzS2sas4bXWNWAqGu12I2Ekk6il0plRjk8j9wcAoAAyKCFaTFUpnJohMKRDrdDy9fqDKE3SaqGbUM4LJYraHrTbbER7ERI6gUKDGKCkbFQcRYLCBABEHDgGF5MQ8AoAXnyAI4sYy8un7KSyeSKLlYFxpLqxJDiKAuKZnLQuYxNXniJBisUcFgW3lpDUcjzGWI7IwIMjQB0s8bi51iDza3X64Ggc5Gk1mi1Wm12zUeR3O13u2Nx4zeuBi307IA

Gudahtt: Good catch, you're right, an intersection type is what was needed here.

Gudahtt: Oh I see, that was the name of the generic parameter, I missed that. That works.

mcmire: Yeah feel free to name it whatever's less confusing :)

mcmire (approved):
Looks good!

src/block-tracker-inspector.ts
Gudahtt: Done here: https://github.com/MetaMask/eth-json-rpc-middleware/pull/183/commits/0c3afb7f0e981e6f62d7d5b27e93051bee7e2089

legobeat (approved):

PR #184, opened 2023-01-31
Add a cache strategy enum

Gudahtt:
An enum has been added to represent the cache strategy that we should use for each method.

### DIFF
--- a/src/block-cache.ts
+++ b/src/block-cache.ts
@@ -6,6 +6,7 @@ import {
   blockTagForPayload,
   cacheTypeForMethod,
   canCache,
+  CacheStrategy,
 } from './utils/cache';
 import type {
   Payload,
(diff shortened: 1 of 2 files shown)


src/block-cache.ts
```diff
@@ -150,10 +151,11 @@ export function createBlockCacheMiddleware({
 
   // create caching strategies
   const blockCache: BlockCacheStrategy = new BlockCacheStrategy();
-  const strategies: Record<string, BlockCacheStrategy> = {
-    perma: blockCache,
-    block: blockCache,
-    fork: blockCache,
+  const strategies: Record<CacheStrategy, BlockCacheStrategy | undefined> = {
```
mcmire, line 154: Does the type need to be specified here? Is it just for explicitness?

mcmire (approved):
One question but looks good regardless.

src/block-cache.ts
Gudahtt: It was pre-existing; I hadn't really considered whether it was needed or not, I just made the required updates to it. This library is full of unnecessary types.

Gudahtt: Looks like when I remove this, the inferred type no longer uses `CacheStrategy` as the keys. It gets reduced to the strings instead. Maybe better to preserve the enum as the key so it's clear that it includes all strategies.

Gudahtt:
Rebased to resolve a conflict

src/block-cache.ts
mcmire: Interesting. No problem.

mcmire (approved):
Looks good!

mcmire (approved):

PR #188, opened 2023-02-03
Update `eth-block-tracker` from v5 to v6

Gudahtt:
The breaking changes are changes to the `BaseBlockTracker` class (which is not used here), and updating the minimum supported Node.js version to v14. Neither affects how this package is used here.


adonesky1 (approved):
LGTM!

PR #190, opened 2023-02-09
Remove `node-fetch`

Gudahtt:
The package `node-fetch` has been removed, instead requiring that `fetch` be provided as an option for the fetch middleware.

This might reduce our bundle size (depending on whether our bundler is smart enough to omit this package already; I'm not sure). It also eliminates a dependency that has been a source of security advisories in the past.

We will probably want to add this or `isomorphic-fetch` back as a `devDependency` when we add tests for this middleware, but so far this middleware is not tested.


mcmire (approved):
LGTM.

PR #191, opened 2023-02-09
Remove return value from `timeout`

Gudahtt:
The `timeout` function no longer returns a timer. This was simplified so that we'd have one less reliance upon a Node.js type/API in this library, as we use this in-browser. That return type wasn't used anyway.

Tests and documentation have been added. The Jest `toNeverResolve` matcher has been migrated here from the core repository for use in the `timeout` tests.

### DIFF
--- a/test/setupAfterEnv.ts
+++ b/test/setupAfterEnv.ts
@@ -1,0 +1,74 @@
+const UNRESOLVED = Symbol('timedOut');
+// Store this in case it gets stubbed later
+const originalSetTimeout = global.setTimeout;
+const TIME_TO_WAIT_UNTIL_UNRESOLVED = 100;
+
+/**
+ * Produces a sort of dummy promise which can be used in conjunction with a
+ * "real" promise to determine whether the "real" promise was ever resolved. If
+ * the promise that is produced by this function resolves first, then the other
+ * one must be unresolved.
+ *
+ * @param duration - How long to wait before resolving the promise returned by
+ * this function.
+ * @returns A promise that resolves to a symbol.
+ */
+const treatUnresolvedAfter = (duration: number): Promise<typeof UNRESOLVED> => {
+  return new Promise((resolve) => {
+    originalSetTimeout(resolve, duration, UNRESOLVED);
+  });
+};
+
+expect.extend({
+  /**
+   * Tests that the given promise is never fulfilled or rejected past a certain
+   * amount of time (which is the default time that Jest tests wait before
+   * timing out as configured in the Jest configuration file).
+   *
+   * Inspired by <https://stackoverflow.com/a/68409467/260771>.
+   *
+   * @param promise - The promise to test.
+   * @returns The result of the matcher.
+   */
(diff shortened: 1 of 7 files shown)


test/setupAfterEnv.ts
```diff
@@ -0,0 +1,74 @@
...
+            }`;
+          },
+          pass: false,
+        };
+  },
+});
+
+// Export something so that TypeScript thinks that we are performing type
+// augmentation
+export {};
```
mcmire, line 74: This line was necessary for type augmentation, but now that that's been extracted to another file, this line is no longer necessary in this file.

types/global.d.ts
```diff
@@ -0,0 +1,13 @@
+declare global {
+  // Using `namespace` here is okay because this is how the Jest types are
+  // defined.
+  namespace jest {
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars, @typescript-eslint/ban-types
+    interface Matchers<R, T = {}> {
```
mcmire, line 6: I believe technically we can drop the `T = {}`, according to the [docs](https://jestjs.io/docs/27.x/expect#expectextendmatchers).

mcmire (approved):
Some nits, but looks good. Clearly I need to read more about how the event loop works in Node :)

test/setupAfterEnv.ts
Gudahtt: Hmm. Still seems to be required by `import/unambiguous`

types/global.d.ts
Gudahtt: Great! One less ESLint ignore comment now as well. Done in https://github.com/MetaMask/eth-json-rpc-middleware/pull/191/commits/e9f4fc5e3c4c5ebeb2ec2602d49a79abf80edfdc

Gudahtt: I have made the equivalent change on core as well: https://github.com/MetaMask/core/pull/1099

test/setupAfterEnv.ts
mcmire: Oh, okay. Then the comment is wrong, because we're not performing type augmentation here. Would it would be more appropriate in this case to put `/* eslint-disable import/unambiguous */` at the top of the file?

Gudahtt: Good point. At the very least we should make the comment more accurate.

I was thinking that ensuring our modules were identifiable as "Unambiguous JavaScript Grammer" was desirable in general, to ensure we don't get surprised when some tool gets confused. I don't know for certain that it would be relevant for us, but it seems reasonably easy to accommodate.

mcmire (approved):

test/setupAfterEnv.ts
Gudahtt: Comment fixed here: https://github.com/MetaMask/eth-json-rpc-middleware/pull/191/commits/66c03e43f328dad1e549b59639ffae7796bcffa8

mcmire (approved):
Looks good!

PR #192, opened 2023-02-10
Remove `btoa`

Gudahtt:
The `btoa` function must now be passed in as a middleware option. This lets us remove an additional dependency, and makes this package less dependent upon Node.js or browser specific APIs.


Gudahtt:
Depends upon #190

mcmire (approved):
Looks good!

PR #193, opened 2023-02-10
Remove unnecessary `suppressUnauthorized` option

Gudahtt:
The `suppressUnauthorized` option has been removed from the type signature of the `getAccounts` parameter accepted by the `wallet` middleware.

This option was added in #116 to fix an extension bug, but it was never used in practice. The linked extension change used the wrong parameter name (`suppressUnauthorizedError`), so this was always unset. We didn't notice because it also wasn't needed to fix the bug; the `resemblesAddress` condition added in that same PR was sufficient.


shanejonas (dismissed):

Gudahtt:
Rebased to resolve a conflict

mcmire (approved):
😎

PR #194, opened 2023-02-11
Remove `providerFromEngine` and `providerFromMiddleware`

Gudahtt:
The two provider construction functions have been removed. These functions are instead availabile from the new package `eth-json-rpc-provider`.


src/providerAsMiddleware.ts
```diff
@@ -28,10 +28,12 @@ export function ethersProviderAsMiddleware(
     // send request to provider
     provider.send(
       req,
-      (err: Error, providerRes: PendingJsonRpcResponse<any>) => {
+      (err: unknown, providerRes: PendingJsonRpcResponse<any>) => {
         // forward any error
         if (err) {
-          return end(err);
+          // TODO: Remove this cast when next major `json-rpc-engine` release is out
+          // The next release changes how errors are propogated.
```
Gudahtt, line 35: For some reason, `json-rpc-engine` was demanding that this be `Error` or  `null`. Yet it's immediately serialized, and `serializeError` accepts `unknown`.

Seems fine to cast it for now because it ought to be accepting `unknown`, unless I'm missing something.

I see that on the main branch of that repo, `end` no longer accepts an error at all, so we'll have to update how this works with the next major release anyway.

mcmire (approved):
Makes sense to me!

joangava (approved):

joangava (approved):

PR #196, opened 2023-02-17
Update `eth-block-tracker` to v7

Gudahtt:
The package `eth-block-tracker` has been updated to v7. This release included two breaking changes: the removal of the `Provider` type, and a change to the expected type of the `provider` constructor option.

We were already using a provider of the expected type, and we only used the removed `Provider` type for some unnecessary casts. Those casts have been removed.


mcmire (approved):
You love to see those typecasts disappear 💥

PR #198, opened 2023-02-21
Add eth_getPlumeSignature RPC handler

rrrliu:
Context: Implementation of a novel deterministic ECDSA signature scheme in MetaMask https://eprint.iacr.org/2022/1255

Add handler for `eth_getPlumeSignature` to unblock https://github.com/MetaMask/metamask-extension/pull/17482, as we'd like to use this new RPC method to generate the signature + other necessary SNARK inputs.

### DIFF
--- a/src/wallet.ts
+++ b/src/wallet.ts
@@ -62,6 +62,10 @@ export interface WalletMiddlewareOptions {
     req: JsonRpcRequest<unknown>,
     version: string,
   ) => Promise<string>;
+  processPlumeSignature?: (
+    msgParams: MessageParams,
+    req: JsonRpcRequest<unknown>,
+  ) => Promise<string>;
 }
 
 export function createWalletMiddleware({
@@ -75,6 +79,7 @@ export function createWalletMiddleware({
   processTypedMessage,
   processTypedMessageV3,
   processTypedMessageV4,
+  processPlumeSignature,
 }: WalletMiddlewareOptions): JsonRpcMiddleware<string, Block> {
   if (!getAccounts) {
     throw new Error('opts.getAccounts is required');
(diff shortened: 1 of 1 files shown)


mcmire (changes requested):
The test coverage for the wallet middleware is lacking, however, we should try to make sure we add tests for additions we make going forward. Can you see if you can do that? Unfortunately we don't have a great example you can work off of, but you can kind of see how this works for existing RPC methods, e.g. `personal_sign`: https://github.com/MetaMask/eth-json-rpc-middleware/blob/ef36cf4ac6b6fb12b2aa0d1d9942c90ecef42b15/src/wallet.test.ts#L283. You'll want to add a new `describe` block to this file and test out the behavior of `eth_getPlumeSignature` (as least as far as you can go). Can you do that? Let me know if you need help.

Gudahtt:
We should ensure that the extension team wishes to accept this change before we include it here. Just wanted to give you a heads up before you do more work on these tests, that it might be better to get buy-in from that team first.

Divide-By-0:
> We should ensure that the extension team wishes to accept this change before we include it here. Just wanted to give you a heads up before you do more work on these tests, that it might be better to get buy-in from that team first.

Great point! We have already discussed with Alex Bea, Erikson, and Olaf to confirm their buy-in, along with our Metamask grant proposal.

shanejonas:
Could you add the method as a PR to our api-specs repo as well: https://github.com/MetaMask/api-specs

rrrliu:
Thank you both for the great feedback!
- @mcmire added some unit tests to `src/wallet.test.ts`
- @shanejonas just opened https://github.com/MetaMask/api-specs/pull/120

PR #223, opened 2023-07-13
devDeps: bump @metamask/eslint-config* to ^12.0.0

legobeat:
Includes:
- #226 
- #229
- #225 
- #227 

- disable subset of new eslint rules
- lint:fix


test/util/helpers.ts
```diff
@@ -111,7 +112,8 @@ export function buildMockParamsWithBlockParamAt(
  * param missing.
  *
  * @param index - The index within the `params` array where the "block" param
- * *would* appear.
+ * would* appear.
+ * @param blockParamIndex
```
mcmire, line 116: Is this left over from the JSDoc rules, since there is no description here?

test/util/helpers.ts
```diff
@@ -85,6 +85,7 @@ export function buildSimpleFinalMiddleware() {
  * endpoint with the given "block" param added at the end.
  *
  * @param index - The index within the `params` array to add the "block" param.
+ * @param blockParamIndex
```
mcmire, line 88: Is this left over from the JSDoc rules, since there is no description here?

mcmire (commented):
Looks good, just had two questions.

test/util/helpers.ts
mcmire: If `lint:fix` added it before you disabled that rule, then I think it makes sense to remove it, but if it's adding it even now then we can keep it. I was just curious.

src/fetch.ts
```diff
@@ -146,6 +145,7 @@ function parseResponse(fetchRes: Response, body: Record<string, Block>): Block {
  * @param options.rpcUrl - The URL to send the request to.
  * @param options.originHttpHeaderKey - If provider, the origin field for each JSON-RPC request
  * will be attached to each outgoing fetch request under this header.
+ * @param options.req
```
mcmire, line 148: Nit: Another instance — what do you think about adding this back when the JSDoc rules are re-enabled?

mcmire (dismissed):
Left another (non-blocking) comment. Looks good.

mcmire (approved):

mcmire (approved):

PR #235, opened 2023-08-08
Retry JSON RPC requests on hitting rate limit

mitinarseny:


mcmire (commented):
Hi @mitinarseny. Would you mind explaining the purpose of this PR — what problem it's intended to solve and how you solved it?

Also, would you mind adding some tests for this middleware so we can ensure it works as intended going forward? Unfortunately there isn't a set standard for tests in this project but you can kind of get a sense from https://github.com/MetaMask/eth-json-rpc-middleware/blob/main/src/inflight-cache.test.ts and https://github.com/MetaMask/eth-json-rpc-middleware/blob/main/src/retryOnEmpty.test.ts#L120 for some inspiration.

PR #254, opened 2023-10-12
Revert error should not be retried

jpuri:
Fixes: https://github.com/MetaMask/MetaMask-planning/issues/1461

This is an issue brought up by blockaid team, in case of revert we should not retry and send back original error.

### DIFF
--- a/src/retryOnEmpty.test.ts
+++ b/src/retryOnEmpty.test.ts
@@ -1,3 +1,4 @@
+import { errorCodes } from '@metamask/rpc-errors';
 import { providerFromEngine } from '@metamask/eth-json-rpc-provider';
 import type { SafeEventEmitterProvider } from '@metamask/eth-json-rpc-provider';
 import type { JsonRpcMiddleware } from '@metamask/json-rpc-engine';
@@ -203,8 +204,8 @@ describe('createRetryOnEmptyMiddleware', () => {
                       id: req.id,
                       jsonrpc: '2.0',
                       error: {
-                        code: -1,
-                        message: 'oops',
+                        code: errorCodes.rpc.invalidInput,
+                        message: 'execution reverted',
                       },
                     };
                   },
@@ -623,6 +624,50 @@ describe('createRetryOnEmptyMiddleware', () => {
           await engine.handle(request);
 
           expect(finalMiddleware).toHaveBeenCalled();
+        },
+      );
+    });
+  });
+
+  describe('when provider return execution revert error', () => {
+    it('returns the same error to caller', async () => {
+      await withTestSetup(
+        {
+          configureMiddleware: ({ provider, blockTracker }) => {
+            return {
+              middlewareUnderTest: createRetryOnEmptyMiddleware({
+                provider,
+                blockTracker,
+              }),
+            };
+          },
+        },
+        async ({ engine, provider }) => {
+          const request: JsonRpcRequest<string[]> = {
+            id: 123,
+            jsonrpc: '2.0',
+            method: 'eth_call',
+            params: buildMockParamsWithoutBlockParamAt(2, '100'),
+          };
+          stubProviderRequests(provider, [
+            buildStubForBlockNumberRequest(),
+            {
+              request,
+              response: () => {
+                throw {
+                  code: errorCodes.rpc.invalidInput,
+                  message: 'execution reverted',
+                } as any;
+              },
+            },
+          ]);
+          const promiseForResponse = engine.handle(request);
+          expect(await promiseForResponse).toMatchObject({
+            error: expect.objectContaining({
+              code: errorCodes.rpc.invalidInput,
+              message: 'execution reverted',
+            }),
+          });
         },
       );
     });
--- a/src/retryOnEmpty.test.ts
+++ b/src/retryOnEmpty.test.ts
@@ -1,8 +1,8 @@
-import { errorCodes } from '@metamask/rpc-errors';
 import { providerFromEngine } from '@metamask/eth-json-rpc-provider';
 import type { SafeEventEmitterProvider } from '@metamask/eth-json-rpc-provider';
 import type { JsonRpcMiddleware } from '@metamask/json-rpc-engine';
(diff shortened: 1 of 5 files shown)


src/retryOnEmpty.ts
```diff
@@ -140,7 +140,10 @@ async function retry(
   for (let index = 0; index < maxRetries; index++) {
     try {
       return await asyncFn();
-    } catch (err) {
+    } catch (err: any) {
+      if (err.code === -32000) {
+        throw err;
```
mcmire, line 145: Makes sense on the surface!

A few comments:

- `eth-json-rpc-middleware` uses our `@metamask/rpc-errors` package which provides an abstraction around possible RPC errors. It looks like there is a constant that represents this error code. I feel like ought to make use of it so that this isn't a "magic number".
- Since this is a `catch`, we don't really know that `err` has a code property — it's really `unknown` and not `any` — so to be safe, we should check for that property. This package also uses `@metamask/utils`, which has a `isValidJsonRpcError` utility function we can use.
- In your PR title, you say that you're handling a revert error. However, looking at `@metamask/rpc-errors`, [-32000 seems to be the code for an "invalid input" error](https://github.com/MetaMask/rpc-errors/blob/9200551fa80ab1f912704a2b346ef10f506878b0/src/error-constants.ts#L4). Therefore, if we truly want to handle a "revert" error, we might want to check the message as well as the code.

Putting this together, we might want to add a new function:

``` typescript
import { isValidJsonRpcError } from "@metamask/utils";
import type { JsonRpcError } from "@metamask/utils";
import { errorCodes } from "@metamask/rpc-errors";

function isExecutionRevertedError(error: unknown): error is JsonRpcError {
  return isValidJsonRpcError(error) &&
    error.code === errorCodes.rpc.invalidInput &&
    error.message === "execution reverted";
}
```

Then, here, we can say:

``` suggestion
    } catch (err: unknown) {
      if (isExecutionRevertedError(err)) {
        throw err;
```

mcmire (commented):
This change makes sense. I also checked the other middleware to make sure there wasn't any other place we needed to do this. The `fetch` middleware also retries requests, but has an allowlist of retriable errors, and this revert error isn't one of them, so we're good on that (these retriable errors, by the way, are what I think the issue you found is referring to).

Also — and this is just a sidebar — I don't really know why `retryOnEmpty` retries _errors_; based on its name it should only retry empty requests. Unfortunately I don't have enough context on this to know what the right thing to do here is, so it's probably best to leave it alone.

In conclusion, your change makes sense, I just had some tweaks.

src/retryOnEmpty.ts
jpuri: I added this improvement.

jpuri:
Hey @mcmire : I updated the PR and also added unit test coverage.

src/retryOnEmpty.test.ts
```diff
@@ -627,6 +628,50 @@ describe('createRetryOnEmptyMiddleware', () => {
...
+          };
+          stubProviderRequests(provider, [
+            buildStubForBlockNumberRequest(),
+            {
+              request,
+              response: () => {
+                throw {
+                  code: errorCodes.rpc.invalidInput,
+                  message: 'execution reverted',
+                } as any;
```
mcmire, line 660: `rpc-errors` can build the errors too. To avoid the type assertion, what do you think about:

``` suggestion
                throw rpcErrors.invalidInput('execution reverted');
```

(note: you'll have to update the `import` at the top)

src/retryOnEmpty.ts
```diff
@@ -140,7 +141,10 @@ async function retry(
   for (let index = 0; index < maxRetries; index++) {
     try {
       return await asyncFn();
-    } catch (err) {
+    } catch (err: any) {
+      if (isExecutionRevertedError(err)) {
+        throw err as unknown;
+      }
```
mcmire, line 147: Instead of asserting that `err` is `unknown`, can you tell TypeScript to use `unknown` as the type of `err` to begin with?

``` suggestion
    } catch (err: unknown) {
      if (isExecutionRevertedError(err)) {
        throw err;
      }

File: test/util/helpers.ts [hunk 4]
```diff
@@ -121,11 +121,12 @@ export function buildMockParamsWithBlockParamAt(
  */
 export function buildMockParamsWithoutBlockParamAt(
```
mcmire, line 122: This function is called "build mock params _without_ block param". So if you want to add a block param at a specific spot, does it make sense to use `blockMockParamWithBlockParamAt` instead of modifying this function?

src/utils/error.test.ts
```diff
@@ -0,0 +1,36 @@
+import { errorCodes } from '@metamask/rpc-errors';
+
+import { isExecutionRevertedError } from './error';
+
+const executionRevertedError = {
+  code: errorCodes.rpc.invalidInput,
+  message: 'execution reverted',
+};
+
+describe('error', () => {
```
mcmire, line 10: Since you're testing `isExecutionRevertedError`, what do you think about calling this after the function name?

``` suggestion
describe('isExecutionRevertedError', () => {

**Review by mcmire — COMMENTED** [review]
Just a few suggestions. Solution looks good aside from this though!

File: src/retryOnEmpty.ts (hunk 3 above)
Comment (jpuri) [reply src/retryOnEmpty.ts]: It is still giving eslint issue if I remove ` as unknown`

File: test/util/helpers.ts (hunk 4 above)
Comment (jpuri) [reply test/util/helpers.ts:122]: I fixed this

**Comment by jpuri** [comment]
Hey @mcmire : thanks a lot for review feedbacks, I addressed all of those.

File: test/util/helpers.ts [hunk 6]
```diff
@@ -121,11 +121,12 @@ export function buildMockParamsWithBlockParamAt(
...
   blockParamIndex: number,
+  value?: string,
```
mcmire, line 124: Do you want to revert these changes? It doesn't appear they are being used any longer.

jpuri: PR is updated

mcmire (approved):
Looks good!

PR #263, opened 2023-10-26
fix(wallet): add back missing txData for signTransaction,sendTransaction

### LINKED ISSUE #262: `txParams`  getting overwritten in wallet middleware
Recently it was discovered that when upgrading the extension to `eth-json-rpc-middleware` 12.0.0, after sending a transaction via the test dapp, the following error shows up in the console for the background:

> Invalid transaction params: must specify "data" for contract deployments, or "to" (and optionally "data") for all other types of transactions.

It turns out that [this commit](https://github.com/MetaMask/eth-json-rpc-middleware/commit/3910dad7b5411ba2d4cf89bc6eb6f9358f04fd6d#diff-5eaa3c61f9e2360b928d6ec7872297c8f9c7c3ba83724ddb9ee9eb937c954308R158) introduced unintentional breaking changes. In `wallet.ts` -> `sendTransaction`, the `txParams` object is getting completely blown away and replaced with an object that has a single `from` property. This breaks the existing functionality around sending transactions. Other functions have similar issues. We should fix `wallet.ts` so that this does not occur.

Oddly, this didn't break any tests. So, we should make sure to add those as well so that this is not an issue going forward.

legobeat:
- Resolves #262 

- [x] Add regression test(s)
  - Should fail on `main` but pass on top of this PR

### DIFF
--- a/src/wallet.test.ts
+++ b/src/wallet.test.ts
@@ -116,6 +116,26 @@ describe('wallet', () => {
         'The requested account and/or method has not been authorized by the user.',
       );
     });
+
+    it('should not override other request params', async () => {
+      const { engine } = createTestSetup();
+      const getAccounts = async () => testAddresses.slice(0, 2);
+      const witnessedTxParams: TransactionParams[] = [];
+      const processTransaction = async (_txParams: TransactionParams) => {
+        witnessedTxParams.push(_txParams);
+        return testTxHash;
+      };
+      engine.push(createWalletMiddleware({ getAccounts, processTransaction }));
+      const txParams = {
+        from: testAddresses[0],
+        to: testAddresses[1],
+      };
+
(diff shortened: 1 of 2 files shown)


mcmire (commented):
Thank you for doing this. I have looked through your previous commit and compared it to this PR, and it looks like these are the only two that needed to changed.

mcmire:
@legobeat Anything I can do to help you with this? I want to unblock the extension team so that they can upgrade this in the extension, so I'm happy to help you add the missing tests.

cryptodev-2s:
@mcmire I will work on the tests Tomorrow

cryptodev-2s:
replaced by https://github.com/MetaMask/eth-json-rpc-middleware/pull/266

mcmire (approved):
Looks great, thank you!

PR #273, opened 2023-12-02
Add `signatureMethod` property to `MessageParams`

hmalik88:
The `signatureMethod` property was introduced because it will be useful for signature insight snaps to determine which signature scheme to provide insights for. Currently this would be for all our supported signing methods, but in the future can be applied to registered signer snaps.


mcmire (approved):
Makes sense to me!

PR #282, opened 2024-01-22
Update message types

hmalik88:
Wallet middleware was incorrectly assigning `TypedMessageParams` as  the type for `eth_signTypedData` message params. Updated the type and tests.


mcmire (dismissed):
Makes sense. (edit: oh, I see the test failures, okay, I'll check back in a bit)

mcmire (approved):

alanaladd21 (commented):
udate

PR #303, opened 2024-05-23
deps: eth-block-tracker@^8.0.0 -> @metamask/eth-block-tracker@^9.0.2

legobeat:
The package `eth-block-tracker` has been renamed to `@metamask/eth-block-tracker`. 

This upgrades to the latest v9, which contains no breaking changes beyond the rename.

Also contains a bugfix for an issue with multiple redundant polling loops.

https://github.com/MetaMask/eth-block-tracker/blob/main/CHANGELOG.md

### Related
- https://github.com/MetaMask/eth-token-tracker/pull/133

### DIFF
--- a/test/util/helpers.ts
+++ b/test/util/helpers.ts
@@ -145,6 +145,7 @@ export function buildStubForBlockNumberRequest(
   return {
     request: {
       method: 'eth_blockNumber',
+      params: [],
     },
     response: (req) => ({
       id: req.id,
--- a/test/util/helpers.ts
+++ b/test/util/helpers.ts
@@ -145,7 +145,6 @@ export function buildStubForBlockNumberRequest(
   return {
     request: {
       method: 'eth_blockNumber',
-      params: [],
     },
     response: (req) => ({
       id: req.id,
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "test:watch": "jest --watch"
   },
   "dependencies": {
-    "@metamask/eth-block-tracker": "^9.0.1",
+    "@metamask/eth-block-tracker": "^9.0.2",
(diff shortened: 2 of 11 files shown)


legobeat:
It seems like there is a regression in `@metamask/eth-block-tracker@9.0.1`, separate from any of its associated dependency changes (determined by lockbumping all other deps before bumping the package itself).

- [Test failing](https://github.com/MetaMask/eth-json-rpc-middleware/actions/runs/9202012657/job/25311162016?pr=303)
- [Failing commit](https://github.com/MetaMask/eth-json-rpc-middleware/pull/303/commits/74f629e0bc5ab417e8b420b19ea27e67b1e10ee8)
- Seems caused by: https://github.com/MetaMask/eth-block-tracker/pull/208

Currently I don't have insights into why, how to fix, or what actual impact is.

test/util/helpers.ts
```diff
@@ -145,7 +145,6 @@ export function buildStubForBlockNumberRequest(
   return {
     request: {
       method: 'eth_blockNumber',
-      params: [],
```
Gudahtt, line 148: Hmm, with v9.0.2 we should be able to restore this. Though I'm not sure whether it matters

Gudahtt (dismissed):
LGTM!

test/util/helpers.ts
legobeat: Good call. It _looks _like it shouldn't matter but I'm better safe than sorry here so putting it back here as well.

legobeat: At least it's good to see it work fine with both variants after `9.0.2`

Gudahtt (approved):

PR #306, opened 2024-05-24
deps: @metamask/eth-json-rpc-engine@^7.1.1->^8.0.1

legobeat:
- deps: `@metamask/eth-json-rpc-provider@^2.1.0->^3.0.2`
- deps: `@metamask/eth-json-rpc-engine@^7.1.1->^8.0.1`
- deps: `@metamask/eth-block-tracker@^9.0.2->^9.0.3`

#### Blocked by
- [x] #305
- [x] https://github.com/MetaMask/eth-block-tracker/pull/239

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "@metamask/eth-block-tracker": "^9.0.2",
     "@metamask/eth-json-rpc-provider": "^3.0.2",
     "@metamask/eth-sig-util": "^7.0.0",
-    "@metamask/json-rpc-engine": "^8.0.1",
+    "@metamask/json-rpc-engine": "^8.0.2",
     "@metamask/rpc-errors": "^6.0.0",
     "@metamask/utils": "^8.1.0",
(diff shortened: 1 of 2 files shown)


dbrans:
Hi @legobeat. Does this change require a version bump to this package? I'm not familiar with how this should work.

legobeat:
> Hi @legobeat. Does this change require a version bump to this package? I'm not familiar with how this should work.

The changelogs for packages with breaking versions:

https://github.com/MetaMask/core/blob/main/packages/json-rpc-engine/CHANGELOG.md

https://github.com/MetaMask/core/blob/main/packages/eth-json-rpc-provider/CHANGELOG.md

From my perspective arbitrary if this goes into `12.1.1` or `12.2.0`, but in any case should not be breaking.

rekmarks (approved):
I don't see anything here that could count as a breaking change for consumers of this package.

mcmire (approved):
Same, LGTM!

Issue #299, opened 2024-03-21
Adapt to EIP-1193 provider changes


mcmire:
After [`SafeEventEmitterProvider` is updated to support EIP-1193](https://github.com/MetaMask/core/issues/4095) and a new version of `@metamask/eth-json-rpc-provider` is released, we should adapt to the changes:

- We should bump `@metamask/eth-json-rpc-provider` to rely on the new changes.
- At that point, calling `sendAsync` will be deprecated; we should use `request` instead.
  - There are two middleware which use `sendAsync`: `retryOnEmpty` and `block-ref`.
  - We also use `sendAsync` in `providerAsMiddleware`.
  - Finally, there are a lot of tests that mock `sendAsync` on the provider. The tests use a helper, `stubProviderRequests` which does this in particular. These references need to be changed.

PR #317, opened 2024-07-16
adapt to eip-1193 provider changes

cryptodev-2s:
After [`SafeEventEmitterProvider` is updated to support EIP-1193](https://github.com/MetaMask/core/issues/4095) and a new version of `@metamask/eth-json-rpc-provider` is released, we should adapt to the changes:

- We should bump `@metamask/eth-json-rpc-provider` to rely on the new changes.
- At that point, calling `sendAsync` will be deprecated; we should use `request` instead.
  - There are two middleware which use `sendAsync`: `retryOnEmpty` and `block-ref`.
  - We also use `sendAsync` in `providerAsMiddleware`.
  - Finally, there are a lot of tests that mock `sendAsync` on the provider. The tests use a helper, `stubProviderRequests` which does this in particular. These references need to be changed. 

* Fixes [#299](https://github.com/MetaMask/eth-json-rpc-middleware/issues/299)

### DIFF
--- a/src/block-ref.test.ts
+++ b/src/block-ref.test.ts
@@ -116,6 +116,7 @@ describe('createBlockRefMiddleware', () => {
                   id: 1,
                   jsonrpc: '2.0',
                   result: 'something',
+                  error: undefined,
                 });
               },
             );
@@ -208,6 +209,7 @@ describe('createBlockRefMiddleware', () => {
                   id: 1,
                   jsonrpc: '2.0',
                   result: 'something',
+                  error: undefined,
                 });
               },
             );
--- a/src/block-ref.test.ts
+++ b/src/block-ref.test.ts
@@ -106,13 +106,17 @@ describe('createBlockRefMiddleware', () => {
                         '0x100',
                       ),
                     },
-                    response: () => 'something',
+                    response: async () => Promise.resolve('something'),
                   }),
                 ]);
 
                 const response = await engine.handle(request);
 
-                expect(response).toBe('something');
+                expect(response).toStrictEqual({
+                  id: 1,
+                  jsonrpc: '2.0',
+                  result: 'something',
+                });
               },
             );
           });
@@ -152,7 +156,7 @@ describe('createBlockRefMiddleware', () => {
                         '0x100',
                       ),
                     },
-                    response: () => 'something',
+                    response: async () => Promise.resolve('something'),
                   }),
                 ]);
 
@@ -194,13 +198,17 @@ describe('createBlockRefMiddleware', () => {
                         '0x100',
                       ),
                     },
-                    response: () => 'something',
+                    response: async () => Promise.resolve('something'),
                   }),
                 ]);
 
                 const response = await engine.handle(request);
 
-                expect(response).toBe('something');
+                expect(response).toStrictEqual({
+                  id: 1,
+                  jsonrpc: '2.0',
+                  result: 'something',
+                });
               },
             );
           });
@@ -237,7 +245,7 @@ describe('createBlockRefMiddleware', () => {
                         '0x100',
                       ),
                     },
-                    response: () => 'something',
+                    response: async () => Promise.resolve('something'),
                   }),
                 ]);
 
--- a/src/block-ref.test.ts
+++ b/src/block-ref.test.ts
@@ -112,12 +112,7 @@ describe('createBlockRefMiddleware', () => {
 
                 const response = await engine.handle(request);
 
-                expect(response).toStrictEqual({
-                  id: 1,
-                  jsonrpc: '2.0',
-                  result: 'something',
-                  error: undefined,
-                });
+                expect(response).toBe('something');
               },
             );
           });
@@ -205,12 +200,7 @@ describe('createBlockRefMiddleware', () => {
 
                 const response = await engine.handle(request);
 
-                expect(response).toStrictEqual({
-                  id: 1,
-                  jsonrpc: '2.0',
-                  result: 'something',
-                  error: undefined,
-                });
+                expect(response).toBe('something');
               },
             );
           });
--- a/src/block-ref.test.ts
+++ b/src/block-ref.test.ts
@@ -106,11 +106,7 @@ describe('createBlockRefMiddleware', () => {
                         '0x100',
                       ),
                     },
-                    response: (req) => ({
-                      id: req.id,
-                      jsonrpc: '2.0' as const,
-                      result: 'something',
-                    }),
+                    response: () => 'something',
                   }),
                 ]);
 
@@ -161,11 +157,7 @@ describe('createBlockRefMiddleware', () => {
                         '0x100',
                       ),
                     },
-                    response: (req) => ({
-                      id: req.id,
-                      jsonrpc: '2.0' as const,
-                      result: 'something',
-                    }),
+                    response: () => 'something',
                   }),
                 ]);
 
@@ -207,11 +199,7 @@ describe('createBlockRefMiddleware', () => {
                         '0x100',
                       ),
                     },
-                    response: (req) => ({
-                      id: req.id,
-                      jsonrpc: '2.0' as const,
-                      result: 'something',
-                    }),
+                    response: () => 'something',
                   }),
                 ]);
 
@@ -259,11 +247,7 @@ describe('createBlockRefMiddleware', () => {
                         '0x100',
                       ),
                     },
-                    response: (req) => ({
-                      id: req.id,
-                      jsonrpc: '2.0' as const,
-                      result: 'something',
-                    }),
+                    response: () => 'something',
                   }),
                 ]);
 
@@ -303,13 +287,13 @@ describe('createBlockRefMiddleware', () => {
                       blockParam,
                     ),
                   };
-                  const sendAsyncSpy = stubProviderRequests(provider, [
+                  const requestSpy = stubProviderRequests(provider, [
                     buildStubForBlockNumberRequest('0x100'),
                   ]);
 
                   await engine.handle(request);
 
-                  expectProviderRequestNotToHaveBeenMade(sendAsyncSpy, request);
+                  expectProviderRequestNotToHaveBeenMade(requestSpy, request);
                 },
               );
             });
@@ -387,13 +371,13 @@ describe('createBlockRefMiddleware', () => {
             method: 'a_non_block_param_method',
             params: ['some value', '0x200'],
           };
-          const sendAsyncSpy = stubProviderRequests(provider, [
+          const requestSpy = stubProviderRequests(provider, [
             buildStubForBlockNumberRequest('0x100'),
           ]);
 
           await engine.handle(request);
 
-          expectProviderRequestNotToHaveBeenMade(sendAsyncSpy, request);
+          expectProviderRequestNotToHaveBeenMade(requestSpy, request);
         },
       );
     });
--- a/src/block-ref.test.ts
+++ b/src/block-ref.test.ts
@@ -10,7 +10,7 @@ import {
   stubProviderRequests,
   buildStubForBlockNumberRequest,
   buildStubForGenericRequest,
-  buildFinalMiddlewareWithDefaultResponse,
+  buildFinalMiddlewareWithDefaultResult,
   buildMockParamsWithoutBlockParamAt,
   expectProviderRequestNotToHaveBeenMade,
 } from '../test/util/helpers';
@@ -106,7 +106,7 @@ describe('createBlockRefMiddleware', () => {
                         '0x100',
                       ),
                     },
-                    response: async () => 'something',
+                    result: async () => 'something',
                   }),
                 ]);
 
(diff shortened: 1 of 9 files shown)


src/block-ref.ts
```diff
@@ -68,12 +63,15 @@ export function createBlockRefMiddleware({
 
     // perform child request
     log('Performing another request %o', childRequest);
-    const childRes: PendingJsonRpcResponse<Block> = await pify(
-      provider.sendAsync,
-    ).call(provider, childRequest);
     // copy child response onto original response
-    res.result = childRes.result;
-    res.error = childRes.error;
```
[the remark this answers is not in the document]
cryptodev-2s: hi @legobeat the EIP-1193 changes require the usage of `request` method which return a promise of the result on success or throws an error in other cases  instead of `sendAsync` which takes a callback that can return both error and result

src/providerAsMiddleware.ts
```diff
@@ -10,19 +10,15 @@ export function providerAsMiddleware(
   provider: SafeEventEmitterProvider,
 ): JsonRpcMiddleware<JsonRpcParams, Json> {
   return (req, res, _next, end) => {
-    // send request to provider
-    provider.sendAsync(
-      req,
-      (err: unknown, providerRes: PendingJsonRpcResponse<any>) => {
-        // forward any error
-        if (err instanceof Error) {
```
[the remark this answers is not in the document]
cryptodev-2s: EIP-1193 request method response return a result without any protocol nor error instead the error is thrown

src/providerAsMiddleware.ts
```diff
@@ -10,19 +10,15 @@ export function providerAsMiddleware(
...
-        }
-        // copy provider response onto original response
-        Object.assign(res, providerRes);
-        return end();
-      },
-    );
+    provider
+      .request(req)
+      .then((result) => {
+        res.result = result;
```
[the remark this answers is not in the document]
cryptodev-2s: yes the request method return a result and not a response object (so protocol data are not returned any more `id`, 'version'...)

src/retryOnEmpty.ts
```diff
@@ -103,41 +98,40 @@ export function createRetryOnEmptyMiddleware({
...
+      const childResponse = await retry(10, async () => {
         log('Performing request %o', childRequest);
-        const attemptResponse: PendingJsonRpcResponse<Block> = await pify(
-          provider.sendAsync,
-        ).call(provider, childRequest);
-        log('Response is %o', attemptResponse);
+        const result = await provider.request<JsonRpcParams, Block>(
+          childRequest,
+        );
+        log('Result is %o', result);
```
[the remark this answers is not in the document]
cryptodev-2s: it's part of the ticket request method return a result and not a response object

src/retryOnEmpty.ts
```diff
@@ -103,41 +98,40 @@ export function createRetryOnEmptyMiddleware({
...
         // verify result
-        if (emptyValues.includes(attemptResponse.result as any)) {
+        if (emptyValues.includes(result as any)) {
           throw new Error(
-            `RetryOnEmptyMiddleware - empty response "${JSON.stringify(
-              attemptResponse,
+            `RetryOnEmptyMiddleware - empty result "${JSON.stringify(
```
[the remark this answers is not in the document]
cryptodev-2s: same as above `request` method returns result and not a response object (as previously sendAsync did)

src/block-ref.ts
```diff
@@ -68,12 +63,17 @@ export function createBlockRefMiddleware({
...
+    try {
+      const childResult = await provider.request<JsonRpcParams, Block>(
+        childRequest,
+      );
+      res.result = childResult;
+      res.error = undefined;
+    } catch (error: any) {
```
mcmire, line 73: Do you need the type annotation here?

``` suggestion
    } catch (error) {
```

If you do, then perhaps this should be `unknown`? `any` doesn't really mean "any" type — it's a shortcut for turning off type-checking altogether.

``` suggestion
    } catch (error: unknown) {
```

src/block-ref.ts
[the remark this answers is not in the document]
mcmire: Correct. The `request` method now guarantees that `response.error` will be thrown: https://github.com/MetaMask/core/blob/e441444f473ffe75257000f0948d1e4590790224/packages/eth-json-rpc-provider/src/safe-event-emitter-provider.ts#L94-L102

src/retryOnEmpty.test.ts
```diff
@@ -725,14 +708,14 @@ function stubGenericRequest<T extends JsonRpcParams, U extends Json>(
...
- * Builds a canned response for a request made to `provider.sendAsync` which
+ * Builds a canned response for a request made to `provider.request` which
  * will error for the first N instances and then succeed on the last instance.
  * Intended to be used in conjunction with `stubProviderRequests`.
  *
  * @param request - The request matcher for the stub.
  * @param numberOfTimesToFail - The number of times the request is expected to
  * be called until it returns a successful response.
- * @param successfulResponse - The response that `provider.sendAsync` will
+ * @param successfulResponse - The response that `provider.request` will
```
mcmire, line 718: Since this is no longer expected to return a response object, should we call this `successfulResult` instead?

src/retryOnEmpty.ts
```diff
@@ -103,41 +98,39 @@ export function createRetryOnEmptyMiddleware({
...
-    );
-    // copy child response onto original response
-    res.result = childResponse.result;
-    res.error = childResponse.error;
+        return result;
+      });
+      log('Copying result %o', childResponse.result);
+      res.result = childResponse;
+      res.error = undefined;
+    } catch (error: any) {
```
mcmire, line 121: Similar comment as above: do we need the type annotation?

``` suggestion
    } catch (error) {
```

Or, alternatively, can we use `unknown`?


``` suggestion
    } catch (error: unknown) {
```

src/providerAsMiddleware.ts
```diff
@@ -10,19 +10,15 @@ export function providerAsMiddleware(
   provider: SafeEventEmitterProvider,
 ): JsonRpcMiddleware<JsonRpcParams, Json> {
   return (req, res, _next, end) => {
```
mcmire, line 12: Hmm, I can't add a suggestion here for this, but can we simplify this to use `createAsyncMiddleware`?

``` typescript
  return createAsyncMiddleware(async (req, res, _next) => {
    res.result = await provider.request(req);
  });
```

src/retryOnEmpty.ts
```diff
@@ -103,41 +98,39 @@ export function createRetryOnEmptyMiddleware({
     // create child request with specific block-ref
     const childRequest = klona(req);
     // attempt child request until non-empty response is received
-    const childResponse: PendingJsonRpcResponse<Block> = await retry(
-      10,
-      async () => {
+    try {
```
mcmire, line 101: This `try` seems to wrap the `retry`. But if we wanted to replicate the existing behavior, would we want to wrap the `request` call instead?

test/util/helpers.ts
```diff
@@ -33,10 +28,7 @@ export interface ProviderRequestStub<
   Result extends Json,
 > {
   request: Partial<JsonRpcRequest<Params>>;
-  response: (
-    request: JsonRpcRequest<Params>,
-    callNumber: number,
-  ) => JsonRpcResponse<Result>;
+  response: (callNumber: number) => Promise<Result>;
```
mcmire, line 31: Since this is no longer expected to be the response object, do we want to call this `result` instead?

test/util/helpers.ts
```diff
@@ -225,22 +213,24 @@ export function stubProviderRequests(
...
-      cb(undefined, stub.response(request, callNumber));
+      const res = stub.response(callNumber);
 
       callNumbersByRequest.set(stub.request, callNumber + 1);
 
       if (!stub.remainAfterUse) {
         remainingStubs.splice(stubIndex, 1);
       }
+
+      return Promise.resolve(res);
```
mcmire, line 224: It looks like `stub.response` is a promise now. Do we want to `await` this? Then we don't need to wrap it in a promise below.

``` suggestion
      const result = await stub.response(callNumber);

      callNumbersByRequest.set(stub.request, callNumber + 1);

      if (!stub.remainAfterUse) {
        remainingStubs.splice(stubIndex, 1);
      }

      return result;

File: src/retryOnEmpty.test.ts [hunk 13]
```diff
@@ -750,19 +733,12 @@ function stubRequestThatFailsThenFinallySucceeds<
...
       if (callNumber <= numberOfTimesToFail) {
-        return {
-          id: req.id,
-          jsonrpc: '2.0',
-          error: {
-            code: -1,
-            message: 'oops',
-          },
-        };
+        throw new JsonRpcError(-1, 'oops');
```
mcmire, line 738: Nit: It would be more idiomatic to use the `custom` method on `providerErrors` (and you can get `providerErrors` from `@metamask/rpc-errors`):

``` suggestion
        throw providerErrors.custom({ code: -1, message: 'oops' });
```

src/block-ref.test.ts
```diff
@@ -106,11 +106,7 @@ describe('createBlockRefMiddleware', () => {
                         '0x100',
                       ),
                     },
-                    response: (req) => ({
-                      id: req.id,
-                      jsonrpc: '2.0' as const,
-                      result: 'something',
-                    }),
+                    response: async () => Promise.resolve('something'),
```
mcmire, line 109: Nit: You don't need to wrap the return value of an `async` function in a promise, it will do that for you. So this should be enough:

``` suggestion
                    response: async () => 'something',
```

(I won't note the other instances of this, but you get the idea)

mcmire (commented):
The overall approach here makes sense to me. Most of my comments below are minor.

mcmire:
@SocketSecurity ignore npm/@metamask/eth-block-tracker@10.1.0

This is our own package.

src/retryOnEmpty.ts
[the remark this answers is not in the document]
cryptodev-2s: Here, we are replacing `sendAsync` with `request`. Previously, `sendAsync` returned a response of type `JsonRpcResponse<Result>`, which looked like `{ id: 'SOME_ID', version: '2', result: Result }`. Now, `request` returns only the `Result`. Therefore, it's important to clarify that we are now returning the result directly, not the response object.

src/providerAsMiddleware.ts
```diff
@@ -9,21 +12,9 @@ import type {
...
-          return end(err);
-        }
-        // copy provider response onto original response
-        Object.assign(res, providerRes);
-        return end();
-      },
-    );
-  };
+  return createAsyncMiddleware(async (req, res) => {
+    res.result = await provider.request(req);
```
mcmire, line 16: Just today, @kanthesha and I re-released the EIP-1193 changes in `@metamask/eth-block-tracker` under [11.0.0](https://github.com/MetaMask/eth-block-tracker/releases/tag/v11.0.0) instead of 10.1.0, as we realized they were breaking. I think we will want to make sure that we publish this change under a major version and mark it as breaking as well, because consumers now need to use a provider that supports the `request` method, i.e., they will need to upgrade `@metamask/eth-json-rpc-provider` to upgrade to this change in this package.

src/block-ref.ts
```diff
@@ -68,12 +64,15 @@ export function createBlockRefMiddleware({
...
-    res.error = childRes.error;
+    // copy child result onto original response
+    try {
+      const childResult = await provider.request<JsonRpcParams, Block>(
+        childRequest,
+      );
+      res.result = childResult;
+    } catch (error) {
+      res.error = serializeError(error);
+    }
```
mcmire, line 75: Hmm, now that I think about this, I don't think we need to catch this error. Because we are inside of an async middleware, if `request` throws, the error will already be caught and included in the response object. So we should just be able to say:

``` suggestion
    // copy child result onto original response
    res.result = await provider.request(childRequest);

File: src/retryOnEmpty.ts [hunk 17]
```diff
@@ -103,41 +98,34 @@ export function createRetryOnEmptyMiddleware({
...
   });
 }
 
-async function retry(
+async function retry<Response>(
   maxRetries: number,
-  asyncFn: () => Promise<PendingJsonRpcResponse<Block>>,
-): Promise<PendingJsonRpcResponse<Block>> {
+  asyncFn: () => Promise<Response>,
+): Promise<Response> {
```
mcmire, line 128: Nit: Should we be more clear about what this is (or, rather, is not)?

``` suggestion
async function retry<Result>(
  maxRetries: number,
  asyncFn: () => Promise<Result>,
): Promise<Result> {

mcmire (commented):
Had some suggestions, but overall, the approach makes sense.

src/block-ref.ts
cryptodev-2s: Yes indeed I removed the catch on retry function but forgot to remove it here

cryptodev-2s:
https://github.com/SocketSecurity ignore npm/@metamask/eth-block-tracker@11.0.0

This is our own package.

mcmire (approved):
Looks good!

PR #318, opened 2024-07-18
Changes in typed signature validation and normalization

jpuri:
The PR addresses these 2 issues:

https://github.com/MetaMask/metamask-extension/issues/25733
https://github.com/MetaMask/MetaMask-planning/issues/2810

### DIFF
--- a/.github/workflows/build-lint-test.yml
+++ b/.github/workflows/build-lint-test.yml
@@ -24,7 +24,7 @@ jobs:
       - prepare
     strategy:
       matrix:
-        node-version: [18.x, 20.x, 22.x]
+        node-version: [18.x, 20.x, 22.5.1]
     steps:
       - uses: actions/checkout@v3
       - name: Use Node.js ${{ matrix.node-version }}
@@ -49,7 +49,7 @@ jobs:
       - prepare
     strategy:
       matrix:
-        node-version: [18.x, 20.x, 22.x]
+        node-version: [18.x, 20.x, 22.5.1]
     steps:
       - uses: actions/checkout@v3
       - name: Use Node.js ${{ matrix.node-version }}
@@ -80,7 +80,7 @@ jobs:
       - prepare
     strategy:
       matrix:
-        node-version: [18.x, 20.x, 22.x]
+        node-version: [18.x, 20.x, 22.5.1]
     steps:
       - uses: actions/checkout@v3
       - name: Use Node.js ${{ matrix.node-version }}
--- a/.github/workflows/build-lint-test.yml
+++ b/.github/workflows/build-lint-test.yml
@@ -24,7 +24,7 @@ jobs:
       - prepare
     strategy:
       matrix:
-        node-version: [18.x, 20.x, 22.5.1]
+        node-version: [18.x, 20.x, 22.x]
     steps:
       - uses: actions/checkout@v3
       - name: Use Node.js ${{ matrix.node-version }}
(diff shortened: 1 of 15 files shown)


src/utils/normalize.ts
```diff
@@ -54,36 +52,14 @@ function parseTypedMessage(data: string) {
...
  * @param address - The address to normalize.
  * @returns The normalized address.
  */
-function normalizeContractAddress(address: string): Hex | string {
-  if (isStrictHexString(address) && isValidHexAddress(address)) {
-    return address;
+function normalizeContractAddress(address: Hex): Hex {
+  if (address.startsWith('0X')) {
+    return `0x${address.slice(2)}`;
   }
```
jpuri, line 63: Normilizing decimal and octal value is removed now, only normalization now is replacing `0X` prefix with `0x`.

mcmire (dismissed):
LGTM!

mcmire:
Oof, I see that CI is failing. This could be either a problem where the cache is not being properly restored, or the incorrect set of files is being saved in the cache. In other repos we've discussed replacing `yarn --immutable --immutable-cache` with `yarn --immutable` in `.github/workflows/build-lint-test.yml`. I'll try to investigate soon, but you could experimenting with that if you're stuck.

cryptodev-2s:
> > Oof, I see that CI is failing. This could be either a problem where the cache is not being properly restored, or the incorrect set of files is being saved in the cache. In other repos we've discussed replacing `yarn --immutable --immutable-cache` with `yarn --immutable` in `.github/workflows/build-lint-test.yml`. I'll try to investigate soon, but you could experimenting with that if you're stuck.
> 
> Saw this in other repos when rerunning previously passing jobs. Also specifically Node 22.
> 
> Node.js 22.5.0 was just released so maybe something [lurking here](https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V22.md#22.5.0)?
> 
> Unless it's simply an intermittent infra issue at GH.



> > Oof, I see that CI is failing. This could be either a problem where the cache is not being properly restored, or the incorrect set of files is being saved in the cache. In other repos we've discussed replacing `yarn --immutable --immutable-cache` with `yarn --immutable` in `.github/workflows/build-lint-test.yml`. I'll try to investigate soon, but you could experimenting with that if you're stuck.
> 
> Saw this in other repos when rerunning previously passing jobs. Also specifically Node 22.
> 
> Node.js 22.5.0 was just released so maybe something [lurking here](https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V22.md#22.5.0)?
> 
> Unless it's simply an intermittent infra issue at GH.

@legobeat a fix for this has just been included in a new node patch [`v22.5.1`](https://github.com/nodejs/node/pull/53935)

mcmire:
I've fixed the Node issue here: https://github.com/MetaMask/eth-json-rpc-middleware/pull/321

matthewwalsh0 (dismissed):

OGPoyraz (approved):

mcmire (dismissed):
Looks good!

.github/workflows/build-lint-test.yml
```diff
@@ -32,7 +32,7 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
           cache: 'yarn'
-      - run: yarn --immutable --immutable-cache
```
cryptodev-2s, line 35: Can we rollback this change ?

cryptodev-2s (changes requested):

.github/workflows/build-lint-test.yml
jpuri: 👍 I updated the PR

cryptodev-2s (approved):
LGTM!

PR #320, opened 2024-07-18
Remove support for `eth_sign`

adonesky1:
## Explanation
Months ago, because of phishing risk, we disabled the `eth_sign` API method by default (users could manually enable it with a preference toggle). Now because of additional risk associated with [potentially malicious EIP-3074 invokers](https://ethereum-magicians.org/t/eip-3074-is-unsafe-unnecessary-puts-user-funds-at-risk-while-fragmenting-ux-liquidity-and-the-wallet-stack/19662) we are fully removing support for this method.

## References

Related: https://github.com/MetaMask/MetaMask-planning/issues/2371

Extension PR: https://github.com/MetaMask/metamask-extension/pull/24756

## Removed
- **BREAKING:** Middleware for `eth_sign` method is removed
 - `createWalletMiddleware` no longer accepts `processEthSignMessage` as an option

### DIFF
--- a/src/wallet.ts
+++ b/src/wallet.ts
@@ -54,10 +54,6 @@ export interface WalletMiddlewareOptions {
     address: string,
     req: JsonRpcRequest,
   ) => Promise<string>;
-  processEthSignMessage?: (
-    msgParams: MessageParams,
-    req: JsonRpcRequest,
-  ) => Promise<string>;
   processPersonalMessage?: (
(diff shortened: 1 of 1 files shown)


mcmire (approved):
Hey, sorry for the delay in reviewing this. It might be nice to add a sub-bullet in the changelog entry which lists which option we are removing from `createWalletMiddleware`. Other than that, though, this looks good.

PR #322, opened 2024-07-22
fix: restore pipeline node version to 22.x

cryptodev-2s:
This PR Restore the pipeline node version to `22.x`. This fix follows the temporary solution implemented in [PR #321](https://github.com/MetaMask/eth-json-rpc-middleware/pull/321), which set the version to `22.5.1`. The temporary fix is no longer needed because the merge of [PR node-versions/#182](https://github.com/actions/node-versions/pull/182) now automatically uses the same version.


Gudahtt (approved):
LGTM!

PR #328, opened 2024-09-04
Request validation should not throw if verifyingContract is not defined in typed signature

jpuri:
https://github.com/MetaMask/metamask-extension/issues/26908

### DIFF
--- a/src/wallet.test.ts
+++ b/src/wallet.test.ts
@@ -432,6 +432,50 @@ describe('wallet', () => {
       const promise = pify(engine.handle).call(engine, payload);
       await expect(promise).rejects.toThrow('Invalid input.');
     });
+
+    it('should not throw if verifyingContract is undefined', async () => {
+      const { engine } = createTestSetup();
+      const getAccounts = async () => testAddresses.slice();
+      const witnessedMsgParams: TypedMessageParams[] = [];
+      const processTypedMessageV3 = async (msgParams: TypedMessageParams) => {
+        witnessedMsgParams.push(msgParams);
+        // Assume testMsgSig is the expected signature result
+        return testMsgSig;
+      };
+
+      engine.push(
+        createWalletMiddleware({ getAccounts, processTypedMessageV3 }),
+      );
+
+      const message = {
+        types: {
+          EIP712Domain: [
+            { name: 'name', type: 'string' },
+            { name: 'version', type: 'string' },
+            { name: 'chainId', type: 'uint256' },
+            { name: 'verifyingContract', type: 'address' },
+          ],
+        },
+        primaryType: 'EIP712Domain',
+        message: {},
+      };
+
+      const stringifiedMessage = JSON.stringify(message);
+
+      const payload = {
+        method: 'eth_signTypedData_v3',
+        params: [testAddresses[0], stringifiedMessage], // Assuming testAddresses[0] is a valid address from your setup
+      };
+
+      const promise = pify(engine.handle).call(engine, payload);
+      const result = await promise;
+      expect(result).toStrictEqual({
+        id: undefined,
+        jsonrpc: undefined,
+        result:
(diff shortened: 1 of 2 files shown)


src/wallet.ts
```diff
@@ -463,10 +463,8 @@ WalletMiddlewareOptions): JsonRpcMiddleware<any, Block> {
  * @param data - The data passed in typedSign request.
  */
 function validateVerifyingContract(data: string) {
-  const {
-    domain: { verifyingContract },
-  } = parseTypedMessage(data);
-  if (!isValidHexAddress(verifyingContract)) {
+  const { domain: { verifyingContract } = {} } = parseTypedMessage(data);
+  if (verifyingContract && !isValidHexAddress(verifyingContract)) {
```
jiexi, line 467: Does it matter if `verifyingContract` is an empty string?

pedronfigueiredo (approved):

src/wallet.ts
jpuri: An empty value will not throw error.

jiexi (approved):

src/wallet.ts
Gudahtt: I think @jiexi meant to ask _should_ it throw an error.

I've just tested with v12.0.6 and it looks like an empty `verifyingContract` was accepted though, so continuing to accept it sounds reasonable for now. It did trigger a blockaid warning on v12.0.6, but no crash, and I was able to sign it.

src/wallet.test.ts
```diff
@@ -524,6 +568,35 @@ describe('wallet', () => {
...
+        createWalletMiddleware({ getAccounts, processTypedMessageV4 }),
+      );
+
+      const payload = {
+        method: 'eth_signTypedData_v4',
+        params: [testAddresses[0], JSON.stringify(getMsgParams())],
+      };
+
+      const promise = pify(engine.handle).call(engine, payload);
+      const result = await promise;
```
Gudahtt, line 592: Nit: Unnecessary deferred Promise:
```suggestion
      const result = await pify(engine.handle).call(engine, payload);
```

src/wallet.test.ts
```diff
@@ -432,6 +432,50 @@ describe('wallet', () => {
       const promise = pify(engine.handle).call(engine, payload);
       await expect(promise).rejects.toThrow('Invalid input.');
     });
+
+    it('should not throw if verifyingContract is undefined', async () => {
+      const { engine } = createTestSetup();
+      const getAccounts = async () => testAddresses.slice();
+      const witnessedMsgParams: TypedMessageParams[] = [];
+      const processTypedMessageV3 = async (msgParams: TypedMessageParams) => {
+        witnessedMsgParams.push(msgParams);
```
Gudahtt, line 441: Nit: This variable seems to serve no purpose:
```suggestion
      const processTypedMessageV3 = async (msgParams: TypedMessageParams) => {
```

src/wallet.test.ts
```diff
@@ -524,6 +568,35 @@ describe('wallet', () => {
       const promise = pify(engine.handle).call(engine, payload);
       await expect(promise).rejects.toThrow('Invalid input.');
     });
+
+    it('should not throw if request is permit with undefined value for verifyingContract address', async () => {
+      const { engine } = createTestSetup();
+      const getAccounts = async () => testAddresses.slice();
+      const witnessedMsgParams: TypedMessageParams[] = [];
+      const processTypedMessageV4 = async (msgParams: TypedMessageParams) => {
+        witnessedMsgParams.push(msgParams);
```
Gudahtt, line 577: Nit: This variable isn't used either
```suggestion
      const processTypedMessageV4 = async (msgParams: TypedMessageParams) => {
```

Gudahtt (approved):
LGTM! Some test cleanup suggested, but the tests were already pretty messy to start with

PR #334, opened 2024-09-17
fix: change types signatures verifyingContract validation to allow 'cosmos' as address

jpuri:
Fixes: https://github.com/MetaMask/metamask-extension/issues/27070

Following the discussion in the thread, I am moving this validation to middleware. https://consensys.slack.com/archives/C02GENU377A/p1725914308614879

Original issue which got a hotfix: https://github.com/MetaMask/metamask-extension/issues/26980

### DIFF
--- a/src/wallet.ts
+++ b/src/wallet.ts
@@ -465,7 +465,7 @@ WalletMiddlewareOptions): JsonRpcMiddleware<any, Block> {
 function validateVerifyingContract(data: string) {
   const { domain: { verifyingContract } = {} } = parseTypedMessage(data);
   // Explicit check for cosmos here has been added to address this issue
-  // https://github.com/MetaMask/metamask-extension/issues/26980
+  // https://github.com/MetaMask/eth-json-rpc-middleware/issues/new
   if (
     verifyingContract &&
     (verifyingContract as string) !== 'cosmos' &&
--- a/src/wallet.ts
+++ b/src/wallet.ts
@@ -464,7 +464,13 @@ WalletMiddlewareOptions): JsonRpcMiddleware<any, Block> {
  */
 function validateVerifyingContract(data: string) {
   const { domain: { verifyingContract } = {} } = parseTypedMessage(data);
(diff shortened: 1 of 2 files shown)


src/wallet.ts
```diff
@@ -464,7 +464,13 @@ WalletMiddlewareOptions): JsonRpcMiddleware<any, Block> {
  */
 function validateVerifyingContract(data: string) {
   const { domain: { verifyingContract } = {} } = parseTypedMessage(data);
-  if (verifyingContract && !isValidHexAddress(verifyingContract)) {
+  // Explicit check for cosmos here has been added to address this issue
+  // https://github.com/MetaMask/metamask-extension/issues/26980
+  if (
+    verifyingContract &&
+    (verifyingContract as string) !== 'cosmos' &&
```
adonesky1, line 471: We're going to introduce this single exception? Doesn't it seem possible there are other cases here? Do we have a plan to deprecate support for this exception and start enforcing this more strictly or? 

cc @MetaMask/wallet-api-platform-engineers

jpuri: Cosmos network DAPPs are using this deprecated SDK which has verifyingContract hard coded to `cosmos` https://github.com/evmos/ethermint/blob/main/ethereum/eip712/domain.go#L29 thus we need to do it to support cosmos network. Once this old library goes out of use we can revert the change.

We do not see more such cases currently.

mikesposito: if I understand correctly we are planning on removing this sometime in the future: if so, to make sure we don't lose track of it, perhaps we can open a new issue and mention it in the code comment instead of using the issue that will be closed by this PR (#26980)?

mikesposito (dismissed):
Looks good to me as a temporary fix, I only have a doubt about how we can state more clearly that this is indeed a temporary fix

src/wallet.ts
jpuri: Good point @mikesposito , I added this issue https://github.com/MetaMask/eth-json-rpc-middleware/issues/337 and updated the comment.

mikesposito (approved):

PR #341, opened 2024-10-02
Fix: update wrong cosmos issue url

cryptodev-2s:
This PR update the cosmos issue url that will be removed in the future


Gudahtt (approved):
LGTM!

PR #350, opened 2024-12-16
fix: Add validation for types sign message primary type

jpuri:
Ref: https://github.com/MetaMask/metamask-extension/issues/22899

Throw error if typed sign request does not have primary type defined.

### DIFF
--- a/src/utils/common.test.ts
+++ b/src/utils/common.test.ts
@@ -4,11 +4,11 @@ describe('CommonUtils', () => {
   describe('stripArrayTypeIfPresent', () => {
     it('remove array brackets from the type if present', () => {
       expect(stripArrayTypeIfPresent('string[]')).toBe('string');
-      expect(stripArrayTypeIfPresent('string []')).toBe('string');
     });
 
     it('return types which are not array without any change', () => {
       expect(stripArrayTypeIfPresent('string')).toBe('string');
+      expect(stripArrayTypeIfPresent('string []')).toBe('string []');
     });
   });
 });
--- a/src/utils/common.test.ts
+++ b/src/utils/common.test.ts
@@ -4,6 +4,7 @@ describe('CommonUtils', () => {
   describe('stripArrayTypeIfPresent', () => {
     it('remove array brackets from the type if present', () => {
       expect(stripArrayTypeIfPresent('string[]')).toBe('string');
+      expect(stripArrayTypeIfPresent('string[5]')).toBe('string');
     });
 
     it('return types which are not array without any change', () => {
--- a/src/utils/common.test.ts
+++ b/src/utils/common.test.ts
@@ -4,6 +4,7 @@ describe('CommonUtils', () => {
   describe('stripArrayTypeIfPresent', () => {
     it('remove array brackets from the type if present', () => {
       expect(stripArrayTypeIfPresent('string[]')).toBe('string');
+      expect(stripArrayTypeIfPresent('string []')).toBe('string');
     });
 
     it('return types which are not array without any change', () => {
--- a/src/utils/common.test.ts
+++ b/src/utils/common.test.ts
@@ -1,0 +1,13 @@
+import { stripArrayTypeIfPresent } from './common';
+
+describe('CommonUtils', () => {
+  describe('stripArrayTypeIfPresent', () => {
+    it('remove array brackets from the type if present', () => {
+      expect(stripArrayTypeIfPresent('string[]')).toBe('string');
+    });
+
+    it('return types which are not array without any change', () => {
+      expect(stripArrayTypeIfPresent('string')).toBe('string');
+    });
+  });
+});
--- a/src/utils/common.ts
+++ b/src/utils/common.ts
@@ -5,8 +5,8 @@
  * @returns Parameter string with array brackets [] removed.
  */
 export const stripArrayTypeIfPresent = (typeString: string) => {
-  if (typeString?.match(/\S\s*\[\]$/u) !== null) {
-    return typeString.replace(/\s*\[\]$/gu, '').trim();
+  if (typeString?.match(/\S\[\]$/u) !== null) {
+    return typeString.replace(/\[\]$/gu, '').trim();
   }
   return typeString;
 };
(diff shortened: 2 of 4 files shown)


src/wallet.ts
```diff
@@ -243,6 +244,7 @@ WalletMiddlewareOptions): JsonRpcMiddleware<any, Block> {
 
     const address = await validateAndNormalizeKeyholder(params[0], req);
     const message = normalizeTypedMessage(params[1]);
+    validatePrimaryType(message);
```
OGPoyraz, line 247: Do we not want to validate this before signing process? Like we did it in the `No PrimaryType defined` example in the test dapp?

If this is a malformed signature and not applying the rules of `EIP-712` I don't think we should even show the request in the extension, must be rejected immediately.

OGPoyraz: For example if we change `TYPED_MESSAGE_SCHEMA` in this package to 

```
{
  "type": "object",
  "properties": {
    "types": {
      "type": "object",
      "additionalProperties": {
        "type": "array",
        "items": {
          "type": "object",
          "properties": {
            "name": { "type": "string" },
            "type": { "type": "string" }
          },
          "required": ["name", "type"]
        }
      }
    },
    "primaryType": { "type": "string" },
    "domain": { "type": "object" },
    "message": { "type": "object" }
  },
  "required": ["types", "primaryType", "domain", "message"],
  "dependencies": {
    "primaryType": {
      "properties": {
        "types": {
          "type": "object",
          "required": {
            "$data": "1/primaryType"
          }
        }
      }
    }
  }
}
```

This schema will be validated before signature is getting added into controller. (in `signature-controller`)

jpuri: Here we are doing same, rejecting before signing process.

src/utils/common.ts
```diff
@@ -0,0 +1,6 @@
+export const stripArrayTypeIfPresent = (typeString: string) => {
+  if (typeString && typeString.match(/\S\[\]$/u) !== null) {
```
Gudahtt, line 2: The linter seems to want you to do something like this:
```suggestion
  if (typeString?.match(/\S\[\]$/u) !== null) {
```

src/wallet.ts
Gudahtt: Indeed, in this PR the validation is happening before the signing process..

Though I do wonder whether we should be using a schema instead of manually validating each expectation we have.

src/utils/common.ts
```diff
@@ -0,0 +1,6 @@
+export const stripArrayTypeIfPresent = (typeString: string) => {
```
Gudahtt, line 1: Nit: Would be nice to have a JSDoc block here as well

src/utils/common.test.ts
```diff
@@ -0,0 +1,13 @@
+import { stripArrayTypeIfPresent } from './common';
+
+describe('CommonUtils', () => {
+  describe('stripArrayTypeIfPresent', () => {
+    it('remove array brackets from the type if present', () => {
```
Gudahtt, line 5: Nit: It looks like `stripArrayTypeIfPresent` will ignore a case like `string []`, where there is a space before `[]`. Perhaps we can add a test case for that as well?

Gudahtt (commented):
Looks good! I can re-review when the lint warning is fixed

src/wallet.ts
OGPoyraz: I am fine with both approaches, just wanted to highlight that schema validation is also happening today. On both approaches, I think it make sense to centralise them in one place, do it on the middleware functions make sense as this PR do

jpuri: I agree @OGPoyraz , but schema validation is currently happening in SignatureController, may be we move it here.

src/utils/common.ts
jpuri: done

src/utils/common.test.ts
jpuri: fixed

OGPoyraz (approved):

src/wallet.ts
OGPoyraz: Exactly, there are multiple tasks in the backlog for that, we will handle that next year

src/utils/common.test.ts
Gudahtt: I think it was correct before. You've now updated it to allow `Type []`, which is unspecified behaviour (the spec only references `Type` or `Type[]`). I didn't mean to suggest you update the regex, just that there was a test case missing

From https://eips.ethereum.org/EIPS/eip-712:
>Arrays are either fixed size or dynamic and denoted by Type[n] or Type[] respectively

src/utils/common.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Function to stripe array brackets if string defining the type has it.
+ *
+ * @param typeString - String defining type from which array brackets are required to be removed.
+ * @returns Parameter string with array brackets [] removed.
+ */
+export const stripArrayTypeIfPresent = (typeString: string) => {
+  if (typeString?.match(/\S\s*\[\]$/u) !== null) {
+    return typeString.replace(/\s*\[\]$/gu, '').trim();
```
Gudahtt, line 9: To address the non-compliant `Type []` case:
```suggestion
  if (typeString?.match(/\S\[\]$/u) !== null) {
    return typeString.replace(/\[\]$/gu, '').trim();
```

Gudahtt (changes requested):
Just one adjustment needed for the `Type []` case

Gudahtt (approved):
LGTM!

src/utils/common.test.ts
jpuri: I updated the code and also changed regex to handle fixed size array like `string[5]`.

OGPoyraz (approved):

src/utils/common.test.ts
Gudahtt: Oh nice, never thought about fixed-sized arrays, and they were right in the quote. Good catch

Gudahtt (approved):

PR #353, opened 2024-12-19
fix: regex match condition for primary type

jpuri:
This PR is in continuation of the previous one: https://github.com/MetaMask/eth-json-rpc-middleware/pull/350

It fixes a condition while validating primary type of types signatures.

### DIFF
--- a/src/utils/common.test.ts
+++ b/src/utils/common.test.ts
@@ -10,6 +10,7 @@ describe('CommonUtils', () => {
     it('return types which are not array without any change', () => {
       expect(stripArrayTypeIfPresent('string')).toBe('string');
       expect(stripArrayTypeIfPresent('string []')).toBe('string []');
+      expect(stripArrayTypeIfPresent(undefined as unknown as string)).toBe(undefined);
     });
   });
 });
--- a/src/utils/common.test.ts
+++ b/src/utils/common.test.ts
@@ -10,7 +10,9 @@ describe('CommonUtils', () => {
     it('return types which are not array without any change', () => {
       expect(stripArrayTypeIfPresent('string')).toBe('string');
       expect(stripArrayTypeIfPresent('string []')).toBe('string []');
-      expect(stripArrayTypeIfPresent(undefined as unknown as string)).toBeUndefined();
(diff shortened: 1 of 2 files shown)


src/utils/common.ts
```diff
@@ -5,7 +5,7 @@
  * @returns Parameter string with array brackets [] removed.
  */
 export const stripArrayTypeIfPresent = (typeString: string) => {
-  if (typeString?.match(/\S\[\d*\]$/u) !== null) {
+  if (typeString?.match(/\S\[\d*\]$/u)) {
```
jpuri, line 8: I found this condition not working as expected when `typeString` was `undefined`.

OGPoyraz (dismissed):

src/utils/common.ts
Gudahtt: Makes sense! Could you add a test case for this?

jpuri: PR is updated with test case.

src/utils/common.test.ts
```diff
@@ -10,6 +10,9 @@ describe('CommonUtils', () => {
     it('return types which are not array without any change', () => {
       expect(stripArrayTypeIfPresent('string')).toBe('string');
       expect(stripArrayTypeIfPresent('string []')).toBe('string []');
+      expect(
+        stripArrayTypeIfPresent(undefined as unknown as string),
```
Gudahtt, line 14: Nit: Better to use `ts-expect-error` here instead, so it's clear this is invalid type is intentional. e.g.

```suggestion
        // @ts-expect-error Intentionally testing invalid input
        stripArrayTypeIfPresent(undefined),
```

Gudahtt (approved):
LGTM!

src/utils/common.test.ts
Gudahtt: Explained further here: https://github.com/MetaMask/contributor-docs/blob/main/docs/typescript.md#use-ts-expect-error-to-force-runtime-execution-of-a-branch-for-validation-or-testing

Gudahtt (approved):

PR #357, opened 2025-01-28
Allow createFetchMiddleware to take an RPC service

### LINKED ISSUE #166: The `fetch` middleware produces an empty response when retries have been exhausted
The `fetch` middleware has a series of errors that it will look for when making the request to the RPC endpoint, and if any of them appear, it will retry the request up to 5 times. If the request still fails after the retries have been exhausted, the middleware will do nothing and return an empty response. The empty response is then caught by `json-rpc-engine`, which produces its own error. Thus, the original error is lost. This is in opposition to the Infura middleware exposed by `eth-json-rpc-middleware`, [which produces a "All retries exhausted" error, along with the original error](https://github.com/MetaMask/eth-json-rpc-infura/blob/276db94947f92e99268f4518f73d625edc766088/src/create-infura-middleware.ts#L107).

### LINKED ISSUE #207: [fetch] Add support for Infura rate limits
See https://github.com/MetaMask/eth-json-rpc-infura/issues/33 for details.

We should add support for Infura rate limits here as well, so that the fetch middleware respects these rate limits when used for custom Infura RPC endpoints (and for any RPC endpoints that behave similarly)

### LINKED ISSUE #209: [fetch] Customizable max retries
Currently the maximum number of retries is hard-coded at 5. We should instead allow this to be configurable, with 5 as the default.

### LINKED ISSUE #211: [fetch] Use exponential backoff when retrying requests
When we encounter a "retriable" error, we should use an exponential backoff to increase the chances of the request succeeding and decrease the load upon the RPC endpoint.

### LINKED ISSUE #356: Use RPC service object from NetworkController in `fetch` middleware
As a part of the [RPC failover project](https://github.com/MetaMask/core/issues/4991), we are adding an `RpcService` class to `@metamask/network-controller`. This class will not only implement the retry and circuit breaker patterns but will also include the same logic as in this package and `@metamask/eth-json-rpc-infura` to retry specific connectivity and response errors.

With this ticket, we will be able to knock out:

- #207 
- #209 
- #211
- #166

mcmire:
`@metamask/network-controller` now contains an `RpcService` class. Using this class not only allows us to remove a lot of code from this package, as it incorporates the vast majority of logic contained in `createFetchMiddleware`, including the retry logic, but it also allows us to use the circuit breaker pattern and the exponential backoff pattern to prevent too many retries if the network is unreliable, and then automatically cut over to a failover node when the network truly goes down.

Closes #356.
Closes #207.
Closes #209.
Closes #211.
Closes #166.

### DIFF
--- a/src/fetch.ts
+++ b/src/fetch.ts
@@ -2,9 +2,10 @@ import type { JsonRpcMiddleware } from '@metamask/json-rpc-engine';
 import { createAsyncMiddleware } from '@metamask/json-rpc-engine';
 import type { JsonRpcError, DataWithOptionalCause } from '@metamask/rpc-errors';
 import { rpcErrors } from '@metamask/rpc-errors';
+import { isJsonRpcFailure } from '@metamask/utils';
 import type { Json, JsonRpcParams, JsonRpcRequest } from '@metamask/utils';
 
-import type { Block } from './types';
+import type { AbstractRpcService, Block } from './types';
 import { timeout } from './utils/timeout';
 
 const RETRIABLE_ERRORS: string[] = [
@@ -21,6 +22,16 @@ const RETRIABLE_ERRORS: string[] = [
 export interface PayloadWithOrigin extends JsonRpcRequest {
   origin?: string;
 }
+
+/**
+ * Like a JSON-RPC request, but includes an optional `origin` property.
+ * This will be included in the request as a header if specified.
+ */
+export type JsonRpcRequestWithOrigin<Params extends JsonRpcParams> =
+  JsonRpcRequest<Params> & {
+    origin?: string;
+  };
+
 interface Request {
   method: string;
   headers: Record<string, string>;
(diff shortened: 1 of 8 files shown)


mcmire:
@SocketSecurity ignore npm/kind-of@6.0.3
@SocketSecurity ignore npm/decamelize-keys@1.1.1
@SocketSecurity ignore npm/@metamask/eth-query@4.0.0
@SocketSecurity ignore npm/@spruceid/siwe-parser@2.1.0
@SocketSecurity ignore npm/@metamask-previews/controller-utils@11.4.5-preview-3abe3688

New author is OK. Either it's us or known members of the NPM community.
Network access is OK. `@metamask/controller-utils` includes several wrappers around `fetch`.

mcmire:
@SocketSecurity ignore npm/@metamask/eth-json-rpc-infura@10.0.0

This is our package.

mcmire:
@SocketSecurity ignore npm/@metamask/controller-utils@11.5.0

This is our package.

src/fetch.ts
```diff
@@ -21,6 +22,16 @@ const RETRIABLE_ERRORS: string[] = [
 export interface PayloadWithOrigin extends JsonRpcRequest {
   origin?: string;
 }
+
+/**
+ * Like a JSON-RPC request, but includes an optional `origin` property.
+ * This will be included in the request as a header if specified.
+ */
+export type JsonRpcRequestWithOrigin<Params extends JsonRpcParams> =
```
mcmire, line 30: Purposefully not reusing `PayloadWithOrigin` above because 1) it's badly named, 2) it doesn't take any type parameters and 3) we need a type and not an interface going forward.

src/fetch.ts
```diff
@@ -32,32 +43,151 @@ interface FetchConfig {
...
+ * @param args.fetch - The `fetch` function; expected to be equivalent to
+ * `window.fetch`.
+ * @param args.rpcUrl - The URL to send the request to.
+ * @param args.originHttpHeaderKey - If provider, the origin field for each
+ * JSON-RPC request will be attached to each outgoing fetch request under this
+ * header.
+ * @returns The fetch middleware.
+ */
+function createFetchMiddlewareWithoutRpcService({
+  btoa: givenBtoa,
```
mcmire, line 178: `btoa` and `fetch` are globals, so we rename them.

src/types.ts
```diff
@@ -23,3 +28,63 @@ export type Block = Record<string, BlockData>;
 export type BlockCache = Record<string, Block>;
 
 export type Cache = Record<number, BlockCache>;
+
+/**
+ * The interface for a service class responsible for making a request to an RPC
+ * endpoint.
+ */
+export type AbstractRpcService = {
```
mcmire, line 36: We could get this from `@metamask/network-controller`, but that would create a circular dependency.

src/fetch.ts
mikesposito: Do you think we should mark `PayloadWithOrigin` as deprecated or do you see it still having some use case in the future?

src/fetch.ts
```diff
@@ -32,32 +43,151 @@ interface FetchConfig {
...
+          params: req.params,
+        },
+        {
+          headers,
+        },
+      );
+
+      // Discard the `id` and `jsonrpc` fields in the response body
+      // (the JSON-RPC engine will fill those in)
+
```
mikesposito, line 152: nit: was this supposed to go a few lines down with this assignment?
```typescript
res.result = jsonRpcResponse.result;
```

mikesposito (dismissed):
Looks great! just a couple of minor comments

src/fetch.ts
mcmire: Good point! I have done so in [12ec424](https://github.com/MetaMask/eth-json-rpc-middleware/pull/357/commits/12ec42419ce293c7370e140756f02d5678d57cab).

src/fetch.ts
mcmire: Yes, good catch. I have moved this closer in [12ec424](https://github.com/MetaMask/eth-json-rpc-middleware/pull/357/commits/12ec42419ce293c7370e140756f02d5678d57cab).

mikesposito (approved):

PR #358, opened 2025-02-03
Bump `@metamask/utils` to ^11.1.0

mcmire:
In a future commit we want to add `@metamask/controller-utils` as a dependency. However, doing so would create type errors because this package relies on a newer version of `@metamask/utils` than `@metamask/eth-json-rpc-middleware` and so there would be two copies in the same dependency tree.

There should be no functional changes to this package. 

Changelog for `@metamask/utils` 11.1.0: https://github.com/MetaMask/utils/blob/main/CHANGELOG.md#1110

Unblocks #356.


cryptodev-2s (approved):
LGTM!

PR #359, opened 2025-02-11
feat: support EIP-5792 methods

matthewwalsh0:
Support the EIP-5792 methods:

- `wallet_sendCalls`
- `wallet_getCallsStatus`
- `wallet_getCapabilities`

Specifically:

- Modularise new method support into separate files under `src/methods`.
  - As opposed to continually extending `src/wallet.ts`.
- Use `@metamask/superstruct` to easily validate request parameters against a  schema / struct.
- Move common validation functions to new `src/validation.ts`.
- Export new types inferred from structs.
  - Specifically as opposed to `*` for easier API diffs.

An additional benefit of using `@metamask/superstruct` is including multiple validation errors within a single RPC error:

```
Error: Invalid params

0 > from - Expected a string matching \`/^0x[0-9a-fA-F]{40}$/\` but received "123"
0 > chainId - Expected a string, but received: 123
0 > calls - Expected an array value, but received: "123"
```

### DIFF
--- a/src/methods/wallet-get-calls-status.test.ts
+++ b/src/methods/wallet-get-calls-status.test.ts
@@ -1,0 +1,114 @@
+import type { JsonRpcRequest, PendingJsonRpcResponse } from '@metamask/utils';
+import { klona } from 'klona';
+
+import type {
+  GetCallsStatusParams,
+  GetCallsStatusResult,
+  GetTransactionReceiptsByBatchIdHook,
+} from './wallet-get-calls-status';
+import { walletGetCallsStatus } from './wallet-get-calls-status';
+
+const ID_MOCK = '1234-5678';
+
+const RECEIPT_MOCK = {
+  logs: [
+    {
+      address: '0x123abc123abc123abc123abc123abc123abc123a',
+      data: '0x123abc',
+      topics: ['0x123abc'],
+    },
+  ],
+  status: '0x1',
+  chainId: '0x1',
+  blockHash: '0x123abc',
+  blockNumber: '0x1',
+  gasUsed: '0x1',
+  transactionHash: '0x123abc',
+};
+
+const REQUEST_MOCK = {
+  params: [ID_MOCK],
+} as unknown as JsonRpcRequest<GetCallsStatusParams>;
+
+describe('wallet_getCallsStatus', () => {
+  let request: JsonRpcRequest<GetCallsStatusParams>;
+  let params: GetCallsStatusParams;
+  let response: PendingJsonRpcResponse<GetCallsStatusResult>;
+  let getTransactionReceiptsByBatchIdMock: jest.MockedFunction<GetTransactionReceiptsByBatchIdHook>;
+
+  async function callMethod() {
+    return walletGetCallsStatus(request, response, {
+      getTransactionReceiptsByBatchId: getTransactionReceiptsByBatchIdMock,
+    });
+  }
+
+  beforeEach(() => {
+    jest.resetAllMocks();
+
+    request = klona(REQUEST_MOCK);
+    params = request.params as GetCallsStatusParams;
+    response = {} as PendingJsonRpcResponse<GetCallsStatusResult>;
+
+    getTransactionReceiptsByBatchIdMock = jest
+      .fn()
+      .mockResolvedValue([RECEIPT_MOCK, RECEIPT_MOCK]);
+  });
+
+  it('calls hook', async () => {
+    await callMethod();
+    expect(getTransactionReceiptsByBatchIdMock).toHaveBeenCalledWith(
+      params[0],
+      request,
+    );
+  });
+
+  it('returns confirmed status if all receipts available', async () => {
+    await callMethod();
+    expect(response.result?.status).toBe('CONFIRMED');
+  });
+
+  it('returns pending status if missing receipts', async () => {
+    getTransactionReceiptsByBatchIdMock = jest
+      .fn()
+      .mockResolvedValue([RECEIPT_MOCK, undefined]);
+
+    await callMethod();
+    expect(response.result?.status).toBe('PENDING');
+  });
+
+  it('returns receipts', async () => {
+    await callMethod();
+
+    expect(response.result?.receipts).toStrictEqual([
+      RECEIPT_MOCK,
+      RECEIPT_MOCK,
+    ]);
+  });
+
+  it('returns null if no receipts', async () => {
+    getTransactionReceiptsByBatchIdMock = jest.fn().mockResolvedValue([]);
+
+    await callMethod();
+    expect(response.result).toBeNull();
+  });
+
+  it('throws if no params', async () => {
+    request.params = undefined;
+
+    await expect(callMethod()).rejects.toMatchInlineSnapshot(`
+            [Error: Invalid params
+
+            Expected an array, but received: undefined]
+          `);
+  });
+
+  it('throws if wrong type', async () => {
+    params[0] = 123 as never;
+
+    await expect(callMethod()).rejects.toMatchInlineSnapshot(`
+            [Error: Invalid params
+
+            0 - Expected a string, but received: 123]
+          `);
+  });
+});
--- a/src/methods/wallet-get-calls-status.test.ts
+++ b/src/methods/wallet-get-calls-status.test.ts
@@ -92,6 +92,12 @@ describe('wallet_getCallsStatus', () => {
     expect(response.result).toBeNull();
   });
 
+  it('throws if no hook', async () => {
+    await expect(
+      walletGetCallsStatus(request, response, {}),
+    ).rejects.toMatchInlineSnapshot(`[Error: Method not supported.]`);
+  });
+
   it('throws if no params', async () => {
     request.params = undefined;
 
@@ -109,6 +115,16 @@ describe('wallet_getCallsStatus', () => {
             [Error: Invalid params
 
             0 - Expected a string, but received: 123]
+          `);
+  });
+
+  it('throws if empty', async () => {
+    params[0] = '';
+
+    await expect(callMethod()).rejects.toMatchInlineSnapshot(`
+            [Error: Invalid params
+
+            0 - Expected a nonempty string but received an empty one]
           `);
   });
 });
--- a/src/methods/wallet-get-calls-status.test.ts
+++ b/src/methods/wallet-get-calls-status.test.ts
@@ -74,6 +74,7 @@ describe('wallet_getCallsStatus', () => {
 
     await callMethod();
     expect(response.result?.status).toBe('PENDING');
+    expect(response.result?.receipts).toBeNull();
   });
 
   it('returns receipts', async () => {
@@ -126,5 +127,19 @@ describe('wallet_getCallsStatus', () => {
 
             0 - Expected a nonempty string but received an empty one]
           `);
+  });
+
+  it('removes excess properties from receipts', async () => {
+    getTransactionReceiptsByBatchIdMock.mockResolvedValue([
+      {
+        ...RECEIPT_MOCK,
+        extra: 'value1',
+        logs: [{ ...RECEIPT_MOCK.logs[0], extra2: 'value2' }],
+      } as never,
+    ]);
+
+    await callMethod();
+
+    expect(response.result?.receipts).toStrictEqual([RECEIPT_MOCK]);
   });
 });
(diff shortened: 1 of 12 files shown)


src/methods/wallet-get-calls-status.ts
```diff
@@ -0,0 +1,85 @@
...
+const GetCallsStatusReceiptStruct = object({
+  logs: array(
+    object({
+      address: HexChecksumAddressStruct,
+      data: StrictHexStruct,
+      topics: array(StrictHexStruct),
+    }),
+  ),
+  status: StrictHexStruct,
+  chainId: optional(StrictHexStruct),
```
jiexi, line 32: is chainId optional in the result? I know we mean for it to be optional in the params of these new methods

src/methods/wallet-get-calls-status.ts
```diff
@@ -0,0 +1,85 @@
...
+    getTransactionReceiptsByBatchId?: GetTransactionReceiptsByBatchIdHook;
+  },
+): Promise<void> {
+  if (!getTransactionReceiptsByBatchId) {
+    throw rpcErrors.methodNotSupported();
+  }
+
+  if (!validateParams(req.params, GetCallsStatusStruct)) {
+    return;
+  }
```
jiexi, line 67: should this throw an error instead?

src/methods/wallet-get-calls-status.ts
```diff
@@ -0,0 +1,85 @@
...
+export type GetCallsStatusReceipt = Infer<typeof GetCallsStatusReceiptStruct>;
+
+export type GetCallsStatusResult = {
+  status: 'PENDING' | 'CONFIRMED';
+  receipts?: GetCallsStatusReceipt[];
+};
+
+export type GetTransactionReceiptsByBatchIdHook = (
+  batchId: string,
+  req: JsonRpcRequest,
```
jiexi, line 49: just curious why the whole req object is needed in the params since it isn't obvious based on the hook name

src/methods/wallet-get-calls-status.test.ts
```diff
@@ -0,0 +1,130 @@
...
+    ]);
+  });
+
+  it('returns null if no receipts', async () => {
+    getTransactionReceiptsByBatchIdMock = jest.fn().mockResolvedValue([]);
+
+    await callMethod();
+    expect(response.result).toBeNull();
+  });
+
```
jiexi, line 94: is it necessary to test the `mask()` call?

src/methods/wallet-get-calls-status.test.ts
```diff
@@ -0,0 +1,130 @@
...
+      request,
+    );
+  });
+
+  it('returns confirmed status if all receipts available', async () => {
+    await callMethod();
+    expect(response.result?.status).toBe('CONFIRMED');
+  });
+
+  it('returns pending status if missing receipts', async () => {
```
jiexi, line 70: should we also test that `receipts` `null` if there are pending?

src/methods/wallet-get-capabilities.ts
```diff
@@ -0,0 +1,45 @@
...
+  PendingJsonRpcResponse,
+} from '@metamask/utils';
+import { HexChecksumAddressStruct } from '@metamask/utils';
+
+import { validateParams } from '../utils/validation';
+
+const GetCapabilitiesStruct = tuple([HexChecksumAddressStruct]);
+
+export type GetCapabilitiesParams = Infer<typeof GetCapabilitiesStruct>;
+export type GetCapabilitiesResult = Record<Hex, Record<string, any>>;
```
jiexi, line 17: i know the specs say this is `any`, but is it actually `any`?

src/methods/wallet-get-capabilities.ts
```diff
@@ -0,0 +1,45 @@
...
+  }: {
+    getCapabilities?: GetCapabilitiesHook;
+  },
+): Promise<void> {
+  if (!getCapabilities) {
+    throw rpcErrors.methodNotSupported();
+  }
+
+  if (!validateParams(req.params, GetCapabilitiesStruct)) {
+    return;
```
jiexi, line 38: should we throw here?

src/methods/wallet-send-calls.ts
```diff
@@ -0,0 +1,80 @@
...
+  res: PendingJsonRpcResponse<Json>,
+  {
+    getAccounts,
+    processSendCalls,
+  }: {
+    getAccounts: (req: JsonRpcRequest) => Promise<string[]>;
+    processSendCalls?: ProcessSendCallsHook;
+  },
+): Promise<void> {
+  if (!processSendCalls) {
```
jiexi, line 59: should this throw is getAccounts is not provided as well?

src/methods/wallet-send-calls.ts
```diff
@@ -0,0 +1,80 @@
...
+import { HexChecksumAddressStruct, StrictHexStruct } from '@metamask/utils';
+
+import {
+  validateAndNormalizeKeyholder,
+  validateParams,
+} from '../utils/validation';
+
+const SendCallsStruct = tuple([
+  object({
+    version: optional(nonempty(string())),
```
jiexi, line 26: this doesn't seem to be optional in the spec

src/methods/wallet-send-calls.ts
```diff
@@ -0,0 +1,80 @@
...
+    getAccounts: (req: JsonRpcRequest) => Promise<string[]>;
+    processSendCalls?: ProcessSendCallsHook;
+  },
+): Promise<void> {
+  if (!processSendCalls) {
+    throw rpcErrors.methodNotSupported();
+  }
+
+  if (!validateParams(req.params, SendCallsStruct)) {
+    // Not possible as throws.
```
jiexi, line 64: ohhh i see. if `validateParams` throws, why do we need it in an if statement in each of these handlers?

src/methods/wallet-send-calls.test.ts
```diff
@@ -0,0 +1,176 @@
...
+    params[0].version = '1.0.0';
+    await callMethod();
+    expect(processSendCallsMock).toHaveBeenCalledWith(params[0], request);
+  });
+
+  it('supports capabilities', async () => {
+    params[0].capabilities = { test: 'value' };
+    await callMethod();
+    expect(processSendCallsMock).toHaveBeenCalledWith(params[0], request);
+  });
```
jiexi, line 78: these seem a bit redundant to me given what `calls hook` tests

src/methods/wallet-send-calls.test.ts
```diff
@@ -0,0 +1,176 @@
...
+    await expect(callMethod()).rejects.toMatchInlineSnapshot(`
+            [Error: Invalid params
+
+            0 > from - Expected a string matching \`/^0x[0-9a-fA-F]{40}$/\` but received "0x123"
+            0 > calls > 0 > to - Expected a string matching \`/^0x[0-9a-fA-F]{40}$/\` but received "0x123"]
+          `);
+  });
+
+  it('throws if from is not in accounts', async () => {
+    getAccountsMock.mockResolvedValueOnce([]);
```
jiexi, line 170: should `validateAndNormalizeKeyholder` be mocked instead?

src/utils/validation.ts
```diff
@@ -0,0 +1,65 @@
...
+export async function validateAndNormalizeKeyholder(
+  address: Hex,
+  req: JsonRpcRequest,
+  { getAccounts }: { getAccounts: (req: JsonRpcRequest) => Promise<string[]> },
+): Promise<Hex> {
+  if (
+    typeof address === 'string' &&
+    address.length > 0 &&
+    resemblesAddress(address)
+  ) {
```
jiexi, line 15: is there some helper in utils we could use here?

src/utils/validation.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  throw rpcErrors.invalidParams({
+    message: `Invalid parameters: must provide an Ethereum address.`,
+  });
+}
+
+export function validateParams<ParamsType>(
+  value: unknown | ParamsType,
+  struct: Struct<ParamsType>,
+): value is ParamsType {
```
jiexi, line 41: seems like you might want the following return type instead:

```suggestion
): asserts value is ParamsType {
```

This would allow you to drop the return value in this helper and also the if statement guards in the handlers that call this

src/utils/validation.ts
jiexi: ahh i see. you're just DRYing this helper up. ehh okay, let's just leave as is then

src/methods/wallet-get-calls-status.ts
matthewwalsh0: This struct is only a means to remove any excess properties.

We don't want to validate the wallet itself so I've made all the properties optional, though I'll follow up to confirm why `chainId` was missing in my tests.

src/methods/wallet-get-calls-status.ts
matthewwalsh0: This is standard in the other hooks, as a way to provide any properties set by other middleware such as `networkClientId`.

src/methods/wallet-get-calls-status.test.ts
matthewwalsh0: Done.

src/methods/wallet-get-calls-status.test.ts
matthewwalsh0: Done.

src/methods/wallet-get-capabilities.ts
matthewwalsh0: Fair point, have changed to `Json` since it's being sent over JSON-RPC so is technically limited.

src/methods/wallet-get-capabilities.ts
matthewwalsh0: Removed in favour of `asserts`.

src/methods/wallet-send-calls.ts
matthewwalsh0: This is validated when `createWalletMiddleware` is called.

src/methods/wallet-send-calls.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export async function walletSendCalls(
+  req: JsonRpcRequest,
+  res: PendingJsonRpcResponse<Json>,
+  {
+    getAccounts,
+    processSendCalls,
+  }: {
+    getAccounts: (req: JsonRpcRequest) => Promise<string[]>;
+    processSendCalls?: ProcessSendCallsHook;
```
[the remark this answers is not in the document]
matthewwalsh0: Yes, as otherwise we conceptually shouldn't need to validate it is provided below.

src/methods/wallet-send-calls.ts
matthewwalsh0: It's not technically required yet as we could assume no version is equivalent to `1.0`.

But I've made it required to be explicit and compliant with the spec.

src/methods/wallet-send-calls.ts
matthewwalsh0: To enforce the type guard, but no longer required with `asserts`.

src/methods/wallet-send-calls.test.ts
matthewwalsh0: A trade-off between test overlap and fragility. Since it was a one liner, I opted to treat the util as an implementation detail and not couple the unit tests to it.

src/utils/validation.ts
matthewwalsh0: Exactly, I wanted to use `isValidHexAddress` but didn't want to impact all the existing usages.

src/utils/validation.ts
matthewwalsh0: Done.

OGPoyraz (approved):

src/methods/wallet-send-calls.test.ts
jiexi: Fair point. I struggle with striking a balance in this area. Cool thanks!

jiexi (approved):

src/utils/validation.test.ts
```diff
@@ -0,0 +1,128 @@
...
+      expect(resemblesAddress(ADDRESS_MOCK)).toBe(true);
+    });
+
+    it('returns false if length not correct', () => {
+      expect(resemblesAddress('0x123')).toBe(false);
+    });
+  });
+
+  describe('validateParams', () => {
+    it('does now throw if superstruct returns no error', () => {
```
cryptodev-2s, line 111: Nit:
Small typo not instead of now

cryptodev-2s (approved):
LGTM, I left a tiny typo fix comment

src/utils/validation.test.ts
matthewwalsh0: Will get in any subsequent PR, thanks!

PR #363, opened 2025-03-12
feat: support updated EIP-5792 specification

matthewwalsh0:
Update the EIP-5792 support following changes to the specification.

Specifically:

- `wallet_sendCalls`
  - Support optional custom `id`.
  - Support optional `capabilities` per call.
  - Support `optional` property in capabilities.
  - Return object containing `id` and optional `capabilities`. 
- `wallet_getCallsStatus`
  - Expect hexadecimal `id`.
  - Add `id`, `version`, and optional `capabilities` to result.
  - Change `status` to number.
  - Provider current codes in `GetCallsStatusCode` enum.
  - Use more generic `GetCallsStatusHook`.
- `wallet_getCapabilities`
  - Support optional `chainIds` argument.

### DIFF
--- a/src/methods/wallet-get-calls-status.test.ts
+++ b/src/methods/wallet-get-calls-status.test.ts
@@ -1,14 +1,18 @@
-import type { JsonRpcRequest, PendingJsonRpcResponse } from '@metamask/utils';
+import type {
+  Hex,
+  JsonRpcRequest,
+  PendingJsonRpcResponse,
+} from '@metamask/utils';
 import { klona } from 'klona';
 
 import type {
+  GetCallsStatusHook,
   GetCallsStatusParams,
   GetCallsStatusResult,
-  GetTransactionReceiptsByBatchIdHook,
 } from './wallet-get-calls-status';
 import { walletGetCallsStatus } from './wallet-get-calls-status';
 
-const ID_MOCK = '1234-5678';
+const ID_MOCK = '0x12345678';
 
 const RECEIPT_MOCK = {
   logs: [
@@ -30,15 +34,23 @@ const REQUEST_MOCK = {
   params: [ID_MOCK],
 } as unknown as JsonRpcRequest<GetCallsStatusParams>;
 
+const RESULT_MOCK = {
+  version: '1.0',
+  id: ID_MOCK,
+  chainId: '0x1',
+  status: 1,
+  receipts: [RECEIPT_MOCK, RECEIPT_MOCK],
+};
+
 describe('wallet_getCallsStatus', () => {
   let request: JsonRpcRequest<GetCallsStatusParams>;
   let params: GetCallsStatusParams;
   let response: PendingJsonRpcResponse<GetCallsStatusResult>;
-  let getTransactionReceiptsByBatchIdMock: jest.MockedFunction<GetTransactionReceiptsByBatchIdHook>;
+  let getCallsStatusMock: jest.MockedFunction<GetCallsStatusHook>;
 
   async function callMethod() {
     return walletGetCallsStatus(request, response, {
-      getTransactionReceiptsByBatchId: getTransactionReceiptsByBatchIdMock,
+      getCallsStatus: getCallsStatusMock,
     });
   }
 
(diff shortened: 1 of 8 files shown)


src/methods/wallet-send-calls.ts
```diff
@@ -21,29 +24,43 @@ import {
...
+    optional: optional(boolean()),
+  }),
+);
+
 const SendCallsStruct = tuple([
   object({
     version: nonempty(string()),
+    id: optional(StrictHexStruct),
     from: HexChecksumAddressStruct,
     chainId: optional(StrictHexStruct),
```
adonesky1, line 39: I know this wasn't changed in this PR but is `chainId` actually optional? Doesn't appear to be in the EIP

matthewwalsh0: I understood this was our own decision as we validate against the dApp selected network, and throw if they don't match. So it's an extra convenience to rely on the dApp selected network.

src/methods/wallet-get-calls-status.ts
```diff
@@ -1,85 +1,69 @@
...
 export type GetCallsStatusParams = Infer<typeof GetCallsStatusStruct>;
-export type GetCallsStatusReceipt = Infer<typeof GetCallsStatusReceiptStruct>;
 
 export type GetCallsStatusResult = {
-  status: 'PENDING' | 'CONFIRMED';
-  receipts?: GetCallsStatusReceipt[];
+  version: string;
+  id: Hex;
+  chainId: Hex;
+  status: number;
```
mcmire, line 30: Should this be:

``` suggestion
  status: (typeof GetCallsStatusCode)[keyof typeof GetCallsStatusCode];
```

src/methods/wallet-get-calls-status.ts
```diff
@@ -1,85 +1,69 @@
...
+  capabilities?: Record<string, Json>;
 };
 
-export type GetTransactionReceiptsByBatchIdHook = (
-  batchId: string,
+export type GetCallsStatusHook = (
+  id: GetCallsStatusParams[0],
   req: JsonRpcRequest,
-) => Promise<GetCallsStatusReceipt[]>;
+) => Promise<GetCallsStatusResult[]>;
```
mcmire, line 49: Should this return an array? I'm not seeing that in the spec, unless I missed it.

src/methods/wallet-get-calls-status.ts
```diff
@@ -1,85 +1,69 @@
...
+  id: Hex;
+  chainId: Hex;
+  status: number;
+  receipts?: {
+    logs: {
+      address: Hex;
+      data: Hex;
+      topics: Hex[];
+    }[];
+    status: Hex;
```
mcmire, line 37: Should this be:

``` suggestion
    status: '0x0' | '0x1';

File: src/methods/wallet-get-capabilities.ts [hunk 5]
```diff
@@ -37,7 +41,8 @@ export async function walletGetCapabilities(
   validateParams(req.params, GetCapabilitiesStruct);
 
   const address = req.params[0];
-  const capabilities = await getCapabilities(address, req);
+  const chainIds = req.params[1];
+  const capabilities = await getCapabilities(address, chainIds, req);
```
jiexi, line 45: does it matter if chainIds is an empty array? I assume either way this case will be handled by the `getCapabilities` hook

src/methods/wallet-send-calls.ts
```diff
@@ -21,29 +24,43 @@ import {
...
-    capabilities: optional(type({})),
+    capabilities: optional(CapabilitiesStruct),
   }),
 ]);
 
 export type SendCallsParams = Infer<typeof SendCallsStruct>;
 export type SendCalls = SendCallsParams[0];
 
+export type SendCallsResult = {
+  id: Hex;
```
jiexi, line 56: should this be `StrictHexStruct`?

src/methods/wallet-send-calls.ts
matthewwalsh0: Changed to required for consistency.

src/methods/wallet-get-calls-status.test.ts
```diff
@@ -119,27 +100,23 @@ describe('wallet_getCallsStatus', () => {
           `);
   });
 
-  it('throws if empty', async () => {
-    params[0] = '';
+  it('throws if not hex', async () => {
```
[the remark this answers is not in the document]
matthewwalsh0: Done.

src/methods/wallet-get-calls-status.test.ts
```diff
@@ -119,27 +100,23 @@ describe('wallet_getCallsStatus', () => {
...
 
-  it('removes excess properties from receipts', async () => {
-    getTransactionReceiptsByBatchIdMock.mockResolvedValue([
-      {
-        ...RECEIPT_MOCK,
-        extra: 'value1',
-        logs: [{ ...RECEIPT_MOCK.logs[0], extra2: 'value2' }],
-      } as never,
-    ]);
+  it('throws if empty', async () => {
```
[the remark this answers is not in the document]
matthewwalsh0: Done.

src/methods/wallet-get-capabilities.test.ts
```diff
@@ -35,7 +37,7 @@ describe('wallet_getCapabilities', () => {
   beforeEach(() => {
     jest.resetAllMocks();
 
-    request = klona(REQUEST_MOCK);
+    request = klona(REQUEST_MOCK) as JsonRpcRequest;
```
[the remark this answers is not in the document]
matthewwalsh0: It is, but we only need the params hence the cast. Using `klona` was just less syntax since it's a nested object, and it was already being used elsewhere.

src/methods/wallet-get-calls-status.ts
matthewwalsh0: The enum is a convenience for the common / current codes, but this was intentionally not coupled to it so the client is still free to return newer or even custom codes in future since the codes are grouped similar to HTTP response.

src/methods/wallet-get-calls-status.ts
matthewwalsh0: Good spot, fixed.

src/methods/wallet-get-capabilities.ts
matthewwalsh0: That should be fine as the client would just treat it as empty. Academic for the time being as we don't support any capabilities.

src/methods/wallet-send-calls.ts
matthewwalsh0: This isn't a struct but a standard type as it's the result so no validation needed.

jiexi (approved):

src/methods/wallet-get-capabilities.test.ts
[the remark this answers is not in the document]
mcmire: Gotcha.

src/methods/wallet-get-calls-status.ts
mcmire: Gotcha, makes sense.

mcmire (approved):
Makes sense!

src/methods/wallet-send-calls.ts
adonesky1: Yeah makes sense

adonesky1 (approved):
LGTM

PR #364, opened 2025-03-13
16.0.0

matthewwalsh0:
See `CHANGELOG.md`.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -993,7 +993,27 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@metamask/eth-json-rpc-middleware@^15.0.1, @metamask/eth-json-rpc-middleware@workspace:.":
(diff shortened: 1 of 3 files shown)


yarn.lock
```diff
@@ -993,7 +993,27 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@metamask/eth-json-rpc-middleware@^15.0.1, @metamask/eth-json-rpc-middleware@workspace:.":
+"@metamask/eth-json-rpc-middleware@npm:^15.0.1":
+  version: 15.3.0
```
mikesposito, line 997: Hmm, interesting. It looks like we have two versions of this package in the lock. Do you have a guess for why this is the case?

matthewwalsh0: I though that was strange too 😄 

Circular dependency via the network controller I believe.

mikesposito (approved):
LGTM!

PR #367, opened 2025-03-31
fetch: Treat non-standard JSON-RPC errors as standard

mcmire:
The [JSON-RPC specification](https://www.jsonrpc.org/specification) says that a JSON-RPC-compliant server must return a response with either a `result` or an `error` field, and if an `error` field is present, then it must be an object with the following fields: `code`, `message`, and `data`. The `isJsonRpcError` from `@metamask/utils` (via the [`JsonRpcError` type][2]) not only checks for the three aforementioned properties, but also allows an additional, optional `stack` property to be present. However, it does not allow any other properties beyond the four declared.

This is a problem because the new implementation of the `fetch` middleware makes use of this function to know how to handle errors, and as a result, non-standard error responses are being treated as successful responses (and the errors themselves are being discarded). This is particularly noticeable when using Ganache as a local RPC server, because it produces such non-standard error responses, such as when a transaction fails.

To correct this bug, this commit brings the error-handling code in the `fetch` middleware closer to the original implementation, and updates the tests to ensure that Ganache errors are treated correctly.

[1]: https://www.jsonrpc.org/specification
[2]: https://github.com/MetaMask/utils/blob/55b22a77e75641c4c8b05eec73982084ac9e7332/src/json.ts#L301

### DIFF
--- a/src/fetch.test.ts
+++ b/src/fetch.test.ts
@@ -84,8 +84,8 @@ describe('createFetchMiddleware', () => {
     );
   });
 
-  describe('if the request to the service returns a successful JSON-RPC response', () => {
-    it('includes the `result` field from the RPC service in its own response', async () => {
+  describe('if the response from the service does not contain an `error` field', () => {
+    it('returns a successful JSON-RPC response containing the value of the `result` field', async () => {
       const rpcService = buildRpcService();
       jest.spyOn(rpcService, 'request').mockResolvedValue({
         id: 1,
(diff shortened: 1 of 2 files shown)


cryptodev-2s (approved):
LGTM!

mcmire:
Tested on mobile as a part of getting https://github.com/MetaMask/metamask-mobile/pull/14273 to pass CI and this seems to work.

PR #369, opened 2025-04-11
chore: update codeowners for wallet request handlers

matthewwalsh0:
Update `.CODEOWNERS` for the wallet request handlers such as `wallet_sendCalls` and `eth_signTypeData`.


cryptodev-2s (approved):
LGTM!

PR #396, opened 2025-08-18
feat: add RPC methods described in (revised) EIP-7715

jeffsmale90:
Adds methods `wallet_requestExecutionPermissions` and `wallet_revokeExecutionPermission`, as defined in this revision of the EIP-7715 specification https://github.com/ethereum/ERCs/pull/1098.

This supports Readable Permissions project, and is related to the following PRs:
- https://github.com/MetaMask/delegation-toolkit/pull/60
- https://github.com/MetaMask/metamask-extension/pull/35193

Note: workflows are failing due to existing problems, fixed by https://github.com/MetaMask/eth-json-rpc-middleware/pull/397

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -28,8 +28,8 @@ export type {
 } from './methods/wallet-request-execution-permissions';
 export type {
   ProcessRevokeExecutionPermissionHook,
-  RevokeExecutionPermissionsRequestParams,
-  RevokeExecutionPermissionsResult,
+  RevokeExecutionPermissionRequestParams,
+  RevokeExecutionPermissionResult,
 } from './methods/wallet-revoke-execution-permission';
 export * from './providerAsMiddleware';
 export * from './retryOnEmpty';
--- a/src/methods/wallet-revoke-execution-permission.test.ts
+++ b/src/methods/wallet-revoke-execution-permission.test.ts
@@ -7,7 +7,7 @@ import { klona } from 'klona';
 
 import type {
   ProcessRevokeExecutionPermissionHook,
-  RevokeExecutionPermissionsRequestParams,
+  RevokeExecutionPermissionRequestParams,
 } from './wallet-revoke-execution-permission';
 import { walletRevokeExecutionPermission } from './wallet-revoke-execution-permission';
 
@@ -17,11 +17,11 @@ const REQUEST_MOCK = {
   params: {
     permissionContext: HEX_MOCK,
   },
-} as unknown as JsonRpcRequest<RevokeExecutionPermissionsRequestParams>;
+} as unknown as JsonRpcRequest<RevokeExecutionPermissionRequestParams>;
 
 describe('wallet_revokeExecutionPermission', () => {
-  let request: JsonRpcRequest<RevokeExecutionPermissionsRequestParams>;
-  let params: RevokeExecutionPermissionsRequestParams;
+  let request: JsonRpcRequest<RevokeExecutionPermissionRequestParams>;
+  let params: RevokeExecutionPermissionRequestParams;
   let response: PendingJsonRpcResponse<Json>;
   let processRevokeExecutionPermissionMock: jest.MockedFunction<ProcessRevokeExecutionPermissionHook>;
 
@@ -35,7 +35,7 @@ describe('wallet_revokeExecutionPermission', () => {
     jest.resetAllMocks();
 
     request = klona(REQUEST_MOCK);
-    params = request.params as RevokeExecutionPermissionsRequestParams;
+    params = request.params as RevokeExecutionPermissionRequestParams;
     response = {} as PendingJsonRpcResponse<Json>;
 
     processRevokeExecutionPermissionMock = jest.fn();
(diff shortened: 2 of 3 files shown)


src/methods/wallet-request-execution-permissions.test.ts
```diff
@@ -0,0 +1,201 @@
...
+    expect(processRequestExecutionPermissionsMock).toHaveBeenCalledWith(
+      params,
+      request,
+    );
+  });
+
+  it('throws if no hook', async () => {
+    await expect(
+      walletRequestExecutionPermissions(request, response, {}),
+    ).rejects.toMatchInlineSnapshot(
```
mcmire, line 132: Nit: Curious, why are we using a snapshot in these tests rather than `.toThrow(new Error(...))`?

jeffsmale90: Yeah, that's a bit silly - I think it was due to the complex error messages. I've replaced it with the more succinct error reason.

src/wallet.ts
```diff
@@ -137,17 +149,30 @@
...
     ),
-    wallet_sendCalls: createAsyncMiddleware(async (params, req) =>
-      walletSendCalls(params, req, { getAccounts, processSendCalls }),
+    wallet_sendCalls: createAsyncMiddleware(async (req, res) =>
+      walletSendCalls(req, res, { getAccounts, processSendCalls }),
     ),
-    wallet_getCallsStatus: createAsyncMiddleware(async (params, req) =>
-      walletGetCallsStatus(params, req, {
+    wallet_getCallsStatus: createAsyncMiddleware(async (req, res) =>
+      walletGetCallsStatus(req, res, {
```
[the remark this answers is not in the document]
jeffsmale90: This is an intentional change - it is a change in name only, and those variables only exist within the scope of this closure.

ffmcgee725 (dismissed):

src/methods/wallet-request-execution-permissions.ts
```diff
@@ -0,0 +1,89 @@
...
+  res: PendingJsonRpcResponse<Json>,
+  {
+    processRequestExecutionPermissions,
+  }: {
+    processRequestExecutionPermissions?: ProcessRequestExecutionPermissionsHook;
+  },
+): Promise<void> {
+  if (!processRequestExecutionPermissions) {
+    throw rpcErrors.resourceNotFound(
+      'wallet_requestExecutionPermissions - no middleware configured',
```
jiexi, line 80: we are throwing methodNotSupported() in the 5792 methods. Thoughts on doing the same here?

https://github.com/MetaMask/eth-json-rpc-middleware/blob/main/src/methods/wallet-get-calls-status.ts#L62C21-L62C39

jiexi (dismissed):
functionally LGTM. I have some concerns regarding spec about if isAdjustmentAllowed should even be included as it ultimately is a proxy for something optional and if something is not strictly required for execution then IMO it should not be included in the request by the dapp

Please ping for re-approval if you feel that my nit above is fair

src/methods/wallet-request-execution-permissions.ts
jeffsmale90: Yeah, that's absolutely the better error to throw.

src/methods/wallet-request-execution-permissions.ts
```diff
@@ -0,0 +1,89 @@
...
+  type Json,
+  type JsonRpcRequest,
+  type PendingJsonRpcResponse,
+  StrictHexStruct,
+} from '@metamask/utils';
+
+import { validateParams } from '../utils/validation';
+
+const PermissionStruct = object({
+  type: string(),
```
jiexi, line 26: do we want to make this an enum and enforce it at this level of the middleware, or is that too high up in the stack / not the primary concern of this middleware?

src/methods/wallet-request-execution-permissions.ts
```diff
@@ -0,0 +1,89 @@
...
+  isAdjustmentAllowed: boolean(),
+  data: record(string(), unknown()),
+});
+
+const RuleStruct = object({
+  type: string(),
```
jiexi, line 32: same

src/index.ts
```diff
@@ -21,6 +21,16 @@ export type {
...
 } from './methods/wallet-send-calls';
+export type {
+  RequestExecutionPermissionsRequestParams,
+  RequestExecutionPermissionsResult,
+  ProcessRequestExecutionPermissionsHook,
+} from './methods/wallet-request-execution-permissions';
+export type {
+  ProcessRevokeExecutionPermissionHook,
+  RevokeExecutionPermissionsRequestParams,
+  RevokeExecutionPermissionsResult,
```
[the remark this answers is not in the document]
jeffsmale90: Request allows for multiple permissions, whereas revoke only supports a single permission. This is maybe a little quirky, but it's part of the specification. We could consider making that change at the specification, but I think we should push forward with this change for now (we are not immediately implementing revocation via RPC, so it would not be a breaking change to external developers).

https://github.com/ethereum/ERCs/pull/1098 defines these two methods.

